PR #1, opened 2019-08-23
Add beam manager

iwko:


test/test_beam_manager.py
```diff
@@ -0,0 +1,264 @@
...
+
+class DataFlowManagerTest(BeamTestCase):
+    def setUp(self):
+        super(DataFlowManagerTest, self).setUp()
+        self.data_flow_manager = create_dataflow_manager(
+            project_id='project',
+            runtime="2019-01-01",
+            dataset_name='test_dataset',
+            dataflow_bucket="bucket",
+            internal_tables={"internal_table": "project.dataset.internal_table"},
```
beczkowb: internal_tables should be a list of table names(manager can construct a full table id from given project_id and dataset_name.

biggerquery/beam_manager.py
```diff
@@ -0,0 +1,205 @@
...
+            shard_name_template='S_N',
+            file_name_suffix='.avro')
+
+
+class TemplatedBeamManager:
+    """
+    Decorator that resolves table names/all kind of variables.
+    """
+
+    def __init__(self, beam_manager, internal_tables, external_tables, run_datetime, extras):
```
beczkowb: change run_datetime to runtime everywhere

biggerquery/beam_manager.py
```diff
@@ -0,0 +1,205 @@
...
+        self.beam_manager = beam_manager
+        self.extras = extras
+        self.internal_tables = internal_tables
+        self.external_tables = external_tables
+        self.run_datetime = run_datetime
+
+    def read_from_big_query(self, dataset, project_id, standard_sql, query):
```
beczkowb: I think that sql is enough, we support only standard sql

biggerquery/beam_manager.py
```diff
@@ -0,0 +1,205 @@
...
+            table_name,
+            schema=schema,
+            write_disposition=BigQueryDisposition.WRITE_TRUNCATE)
+
+    def _legacy_table_name(self, table_name):
+        spl = table_name.split('.')
+        return spl[0] + ':' + '.'.join(spl[1:len(spl)])
+
+    @property
+    def run_datetime(self):
```
beczkowb: run_datetime -> run_datetime_str, I know its bad, but let's be consistent with the bigquery dataset manager

biggerquery/beam_manager.py
```diff
@@ -0,0 +1,205 @@
...
+            options.view_as(StandardOptions).runner = 'DataflowRunner'
+        else:
+            options.view_as(StandardOptions).runner = 'DirectRunner'
+        self.beam_manager.beam_manager.pipeline = beam.Pipeline(options=options)
+        return self.beam_manager.beam_manager.pipeline
+
+    def _job_name(self, day, job_name):
+        if 'job_suffix' in self.beam_manager.extras:
+            import re
+            table_prefix_slug = re.sub(r'[^a-z0-9]+', '-', self.beam_manager.extras['job_suffix']).strip('-')
```
beczkowb: get rid of table_prefix

biggerquery/beam_manager.py
```diff
@@ -0,0 +1,205 @@
...
+def _throw_on_none(param, param_name):
+    if param is None:
+        raise ValueError("%s must be defined" % param_name)
+
+
+def create_dataflow_manager(
+        project_id,
+        runtime,
+        dataset_name,
+        dataflow_bucket,
```
beczkowb: internal_tables and external_tables are optional so they should be None by default. I think that the machine_type parameter should has default value to(standard machine).

iwko: external_tables is optional, right. But without internal_tables it doesnt make sense to use this tool

test/test_beam_manager.py
iwko: Look now please.

biggerquery/beam_manager.py
iwko: Look now please.

biggerquery/beam_manager.py
iwko: Look now please.

biggerquery/beam_manager.py
iwko: Look now please.

biggerquery/beam_manager.py
iwko: Look now please.

biggerquery/beam_manager.py
iwko: Look now please.

beczkowb: You can write full table names without using aliases, so it make sense.

iwko: Oh, right

iwko: Look now please

iwko: Look now please

PR #10, opened 2019-11-05
adding labels added

ewebed:


iwko:
fastai dependency is missing https://travis-ci.com/allegro/biggerquery/builds/134986362

e2e/test_dataset_manager.py
```diff
@@ -41,6 +48,25 @@ def tearDown(self):
         self.dataset_manager.remove_dataset()
 
 
+class PartitionedDatasetManagerPropertiesTestCase(DatasetManagerTestCase):
```
iwko: Those tests are not necessary and are not e2e

test/test_user_commons/test_labels.py
```diff
@@ -0,0 +1,70 @@
+from unittest import TestCase
```
iwko: you should create e2e test in `test_dataset_manager` and test on real table in bigquery

e2e/test_dataset_manager.py
```diff
@@ -18,10 +19,16 @@ class DatasetManagerTestCase(TestCase):
...
     def setUp(self):
+        self.dataset_uuid = str(uuid.uuid4()).replace('-', '')
+        self.internal_tables = ['fake_target_table', 'partitioned_fake_target_table', 'loaded_table']
+        self.external_tables = {'some_external': 'table'}
+
         self.test_dataset_id, self.dataset_manager = create_dataset_manager(
             config.PROJECT_ID,
             self.TEST_PARTITION,
-            internal_tables=['fake_target_table', 'partitioned_fake_target_table', 'loaded_table'])
+            dataset_name=self.dataset_uuid,
```
iwko: empty line

biggerquery/user_commons/labels.py
```diff
@@ -0,0 +1,14 @@
+from biggerquery.interactive import interactive_component
```
iwko: Why not in `interactive.py`?

test/test_user_commons/test_labels.py
```diff
@@ -0,0 +1,70 @@
...
+        self.test_should_execute_update_table_with_proper_parameters(add_label._standard_component)
+
+    def test_should_return_raw_component_when_ds_is_not_provided(self):
+        # given
+        add_label = add_label_component(TEST_TABLE_NAME, TEST_LABELS)
+
+        # expect
+        self.test_should_execute_update_table_with_proper_parameters(add_label)
+
+    def test_should_execute_update_table_with_proper_parameters(self, add_label=None):
```
iwko: You should not use test case as method

biggerquery/user_commons/labels.py
iwko: Why it is standalone file? Why you didnt add it to dataset_manager?

test/test_user_commons/test_labels.py
beczkowb: Why not?

biggerquery/user_commons/labels.py
beczkowb: Because there is no reason for making the dataset_manager even bigger. I think we should go in a more modular direction, instead of creating one huge class.

test/test_user_commons/test_labels.py
iwko: It shadows test case. In this case we should create two tests `test_should_execute_update_table_with_add_label` and `test_should_execute_update_table_without_add_label`

e2e/test_dataset_manager.py
beczkowb: e2e tests are slow and hard to run(you need a project). Unit tests are fine. Maybe @ewebed can add one e2e test, but I'm not sure if that a good direction.

biggerquery/user_commons/labels.py
iwko: 👍

e2e/test_dataset_manager.py
iwko: Ok, but those tests should not be in /e2e directory. But imo we need e2e test

PR #13, opened 2019-11-21
Add possibility to pass all dag's parameters during converting from workflow

faderskd:


biggerquery/deployment.py
```diff
@@ -19,7 +19,7 @@ def job_callable(**kwargs):
     return job_callable
 
 
-def workflow_to_dag(workflow, start_from, dag_id):
+def workflow_to_dag(workflow, start_from, dag_id, **dag_kwargs):
```
beczkowb: Adding the kwargs here is not enough, you need to allow the user to pass the kwargs to the Workflow class and then use the kwargs inside workflow_to_dag

faderskd: done

beczkowb: Did you test it using Cloud Composer?

faderskd: not yet, today will be tested

faderskd: after that it will be ready to merge I guess

PR #14, opened 2019-11-22
Monitoring

beczkowb:


biggerquery/monitoring.py
```diff
@@ -0,0 +1,164 @@
...
+                "value": {
+                    "int64Value": 1
+                }
+            }
+        ]
+    }
+
+
+def increment_counter(client, monitoring_config, metric_type, job_id):
+    start = get_now_rfc3339()
```
faderskd: `start` and `end` are actually the same times

biggerquery/monitoring.py
```diff
@@ -0,0 +1,164 @@
...
+
+def increment_job_failure_count(monitoring_config, job_id):
+    try:
+        client = api_client()
+        if not metric_exists(client, monitoring_config.project_resource, BIGGERQUERY_JOB_FAILURE_METRIC_TYPE):
+            api_create_metric(client, monitoring_config.project_resource, BIGGERQUERY_JOB_FAILURE_METRIC)
+            wait_for_metric(client, monitoring_config.project_resource, BIGGERQUERY_JOB_FAILURE_METRIC_TYPE)
+        increment_counter(client, monitoring_config, BIGGERQUERY_JOB_FAILURE_METRIC_TYPE, job_id)
+    except Exception as e:
+        raise MetricError('Cannot increment job failure count: ' + str(e))
```
faderskd: Do you have here full stack trace or only small portion of valuable message ?

test/test_monitoring.py
```diff
@@ -0,0 +1,227 @@
...
+    def test_should_raise_exception_when_no_retries_left(self, metric_exists_mock):
+        # given
+        metric_exists_mock.return_value = False
+
+        # expect
+        with self.assertRaises(monitoring.MetricError):
+            monitoring.wait_for_metric('client', 'project_resource', 'metric_type')
+
+
+class CreateTimeseriesData(TestCase):
```
faderskd: [Not important] Why you need this test ? If it properly builds data then `increment_counter()` function will properly work. I mean if we should test bigger parts of code as a whole except every small part independently ;)

beczkowb: That's unit testing. If you assert created timeseries inside increment_counter test, you test 2 things in the same test(that's not unit testing, you check if increment_counter works assuming that create timeseries should work). Whether I should write the integration test is a different question :)

biggerquery/monitoring.py
beczkowb: good point

biggerquery/monitoring.py
beczkowb: What do you mean by the same?. We need the start and the end because that is how you create meter things in the stackdriver.

test/test_monitoring.py
faderskd: Ok, that's the matter of approach. Sometimes it's better just to test bigger parts which you treat like a black box even in unit tests. Nevermind.

biggerquery/monitoring.py
```diff
@@ -0,0 +1,164 @@
...
+
+
+def meter_job_run_failures(job, monitoring_config):
+    original_run = job.run
+
+    def metered_run(runtime):
+        try:
+            return original_run(runtime)
+        except Exception as e:
+            logger.error(str(e))
```
faderskd: same as above ;)

biggerquery/monitoring.py
faderskd: Oh, ok understand. So you need only difference in time event if it's a matter of milliseconds

beczkowb: Yes, passing the same values will throw an error.

biggerquery/monitoring.py
beczkowb: done

biggerquery/monitoring.py
beczkowb: done

PR #15, opened 2019-12-25
Add possibility to define job graphs

faderskd:


biggerquery/deployment.py
```diff
@@ -5,6 +5,9 @@
 import os
 import shutil
 
+from airflow import models
```
beczkowb: Hide it inside the functions that use it

biggerquery/workflow.py
```diff
@@ -7,15 +9,164 @@ def __init__(self,
...
+    def run(self, runtime):
+        self.job.run(runtime=runtime)
+
+    def __hash__(self):
+        return hash(self.name)
+
+    def __eq__(self, other):
+        return self.name == other.name
+
+    def __str__(self):
```
beczkowb: I suggest implementing __repr__ instead

biggerquery/workflow.py
```diff
@@ -7,15 +9,164 @@ def __init__(self,
...
+
+    def call_on_graph_nodes(self, consumer):
+        self.definition.call_on_graph_nodes(consumer)
+
+    def _parse_definition(self, definition):
+        if isinstance(definition, list):
+            return Definition(self._map_to_workflow_jobs(definition))
+        if isinstance(definition, Definition):
+            return definition
+        raise ValueError(f"Invalid argument {definition}")
```
beczkowb: We might be compatible with python3.5 for a reason :), I suggest removing 3.6 string formatting

biggerquery/workflow.py
```diff
@@ -7,15 +9,164 @@ def __init__(self,
...
+
+    def _build_graph(self, jobs):
+        if isinstance(jobs, list):
+            job_graph = self._convert_list_to_graph(jobs)
+        elif isinstance(jobs, dict):
+            job_graph = jobs
+        else:
+            raise ValueError("Job graph has to be dict or list")
+
+        JobGraphValidator(job_graph).validate()
```
beczkowb: Why not just a function?

requirements/base.txt
```diff
@@ -3,4 +3,5 @@ pandas>=0.23,<=0.24
 google-cloud-storage>=1.16,<=1.18
 mock>=3.0.5
 pyarrow>=0.14
-pydata-google-auth>=0.1.3
\ No newline at end of file
+pydata-google-auth>=0.1.3
+apache-airflow>=1.10.6
```
beczkowb: We don't want airflow as a base requirements, add it as an additional requirement or don't add it at all(it is used only when deployed on airflow or when unit testing our code).

beczkowb:
Can you write a usage example in README?

biggerquery/workflow.py
beczkowb: Ok, I remember now. It's about the python version on Jupyter machines in GCP(it used to be 3.5, I don't know what's the current version).

biggerquery/workflow.py
```diff
@@ -7,15 +9,164 @@ def __init__(self,
...
+        return job_graph
+
+    @staticmethod
+    def _convert_list_to_graph(job_list):
+        graph_as_dict = OrderedDict()
+        if len(job_list) == 1:
+            graph_as_dict[job_list[0]] = []
+        else:
+            for i in range(1, len(job_list)):
+                graph_as_dict[job_list[i - 1]] = [job_list[i]]
```
iwko: What's going on here?

biggerquery/deployment.py
```diff
@@ -58,12 +79,9 @@ def build_dag_file(workflow_import_path,
 import biggerquery as bgq
 from {workflow_import_path} import {workflow_name} as workflow
 
-dag_args, tasks = bgq.workflow_to_dag(workflow, '{start_from}', '{dag_id}')
+#dag = models.DAG(**dag_args)
```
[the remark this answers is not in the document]
faderskd: Unfortunately this line is required (even as a comment). Airflow searches for line like this and when it finds it, it starts to search for dag objects in python module. Without this line airflow ignores generated module.

biggerquery/workflow.py
```diff
@@ -7,15 +9,164 @@ def __init__(self,
...
+
+class WorkflowJob:
+    def __init__(self, job, name):
+        self.job = job
+        self.name = name
+
+    def run(self, runtime):
+        self.job.run(runtime=runtime)
+
+    def __hash__(self):
```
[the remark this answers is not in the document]
faderskd: yep, we use this object inside workflow graphs represented as dictionaries

biggerquery/workflow.py
```diff
@@ -7,15 +9,164 @@ def __init__(self,
...
+        return hash(self.name)
+
+    def __eq__(self, other):
+        return self.name == other.name
+
+    def __str__(self):
+        return "WorkflowJob{job=..., name=%s}" % self.name
+
+
+class Definition:
```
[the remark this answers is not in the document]
faderskd: it is already inside workflow module so the context is known ;)

biggerquery/workflow.py
faderskd: Why as a function :) ? I wanted to clearly separate logic responsible for validation and finding job order running. There are totally different

biggerquery/workflow.py
faderskd: It is the part of the code necessary to preserve backward compatibility. Currently we define workflows by passing list of jobs except graph. New algorithm for running jobs (or converting them to airflow dag) operate on graphs internally. To be able to adjust the old interface to the new algorithm we have to convert list of jobs to graph.

biggerquery/workflow.py
beczkowb: I mean function, not method- validate_job_graph(job_graph). That's not critical, but more pythonic I think.

faderskd: Yep I know but I wanted to clearly separate this block of code and do not move it to another module

iwko:
usage example in README.MD is still missing

PR #17, opened 2020-04-05
Function for importing biggerquery configuration from path

faderskd:


iwko:
CI fails

iwko:
override method test is missing

biggerquery/cli/config_utlis.py
```diff
@@ -0,0 +1,24 @@
...
+
+    if not hasattr(imported_module, object_name):
+        raise BiggerQueryConfigNotFound("Object '%s' not found in module %s" % (object_name, module_path))
+    maybe_biggerquery_config = getattr(imported_module, object_name)
+    if not isinstance(maybe_biggerquery_config, BiggerQueryConfig):
+        raise InvalidConfig("Object '%s' is not BiggerQueryConfig" % config_path)
+    return maybe_biggerquery_config
+
+
+class BiggerQueryConfigNotFound(Exception):
```
beczkowb, line 19: ValueError with proper message is enough. Let's avoid making new exceptions, especially if there is no except for it.

biggerquery/cli/config_utlis.py
```diff
@@ -0,0 +1,24 @@
...
+    pass
+
+
+class InvalidConfig(Exception):
```
beczkowb, line 23: ValueError

biggerquery/cli/config_utlis.py
```diff
@@ -0,0 +1,24 @@
...
+
+def import_config(config_path):
+    object_name = config_path.split('.')[-1]
+    module_path = '.'.join(config_path.split('.')[:-1])
+    imported_module = import_module(module_path)
+
+    if not hasattr(imported_module, object_name):
+        raise BiggerQueryConfigNotFound("Object '%s' not found in module %s" % (object_name, module_path))
+    maybe_biggerquery_config = getattr(imported_module, object_name)
+    if not isinstance(maybe_biggerquery_config, BiggerQueryConfig):
```
beczkowb, line 14: I think it's not a good idea. It makes harder for a user to create their own configuration.

For example, I personally use duck typing when I use google oauth libraries by creating fake token classes, and I'm happy as a user that they don't check types.

test/test_cli/configs/another_configs/config_for_testing.py
```diff
@@ -0,0 +1,3 @@
+from biggerquery.configuration import BiggerQueryConfig
+
+test_config = BiggerQueryConfig('example_test_configuration')
```
beczkowb, line 3: can you add relative import here to test if it still works?

faderskd: i added this import in another file - /configs/config_for_testing.py (it's better place for relative import to add)

biggerquery/cli/config_utlis.py
faderskd: done

biggerquery/cli/config_utlis.py
faderskd: done

biggerquery/configuration.py
```diff
@@ -6,8 +6,16 @@
 from .utils import unzip_file_and_save_outside_zip_as_tmp_file
 
 
-class DatasetConfig(object):
+class BiggerQueryConfig(object):
```
beczkowb, line 9: Maybe just Config? Similarly to spring @Configuration (not SpringConfiguration).

PR #21, opened 2020-07-16
Add cli run

MarcinMoskala:


biggerquery/cli.py
```diff
@@ -0,0 +1,213 @@
...
+from datetime import datetime
+from importlib import import_module
+from pathlib import Path
+from types import ModuleType
+
+import biggerquery as bgq
+from typing import Optional
+
+
+def resolve(path: Path):
```
beczkowb, line 14: add return type (-> str), I might be useful for Pycharm

biggerquery/cli.py
```diff
@@ -0,0 +1,213 @@
...
+    """
+    full_module_file_path = resolve(module_dir / module_file)
+    full_module_file_path = full_module_file_path.replace(resolve(root_package.parent), '')
+    return full_module_file_path \
+               .replace(os.sep, '.')[1:] \
+        .replace('.py', '') \
+        .replace('.__init__', '')
+
+
+def walk_module_paths(root_package: Path) -> [str]:
```
beczkowb, line 46: Are you sure that the return type shouldn't be -> typing.List(str)?

test/test_cli.py
```diff
@@ -0,0 +1,140 @@
+from os.path import realpath
+from pathlib import Path
+from unittest import TestCase
+import biggerquery as bgq
+from biggerquery.cli import *
+
+TEST_MODULE_SRC = 'test_module'
+
+
+class CacheBuilderTestCase(TestCase):
```
beczkowb, line 10: CacheBuilder?

test/test_cli.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def test_walk_module_files(self):
```
beczkowb, line 12: I suggest more descriptive test names, in this case maybe "test_should_walk_through_all_modules_inside_package_tree"

test/test_cli.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def test_walk_modules(self):
+        # given
+        path = Path(TEST_MODULE_SRC)
+
+        # when
+        res = walk_modules(path)
+
+        # then
+        res = list(res)
```
beczkowb, line 51: In case of a longer, non-obvious # then section, I suggest adding some comments that explain what is going on or wrapping assertions with meaningful function

test/test_cli.py
MarcinMoskala: I can change it this way, but for me it is easier to understand test when the name of the tested method is included, and mosts tests here are each designed to test one single method

biggerquery/cli.py
MarcinMoskala: U R ->

PR #24, opened 2020-07-24
Test cli run

MarcinMoskala:


test/test_cli.py
```diff
@@ -136,3 +139,66 @@ def test_should_find_root_package_when_root_used_project_name_used(self):
 
         # then
         self.assertEqual(test_module_src, resolve(res))
+
+    def test_should_run_workflow(self):
```
beczkowb, line 143: When I read this test I wonder what's the difference between each workflow you run. Do they represent different situations?

test/test_cli.py
```diff
@@ -136,3 +139,66 @@ def test_should_find_root_package_when_root_used_project_name_used(self):
...
+        # then
+        self.assert_started_jobs(['J_ID_3', 'J_ID_4'])
+
+        # when
+        cli_run(root_package, workflow_id="ID_4")
+
+        # then
+        self.assert_started_jobs(['J_ID_3', 'J_ID_4', 'J_ID_5'])
+
+    def test_should_run_workflow_multiple_times(self):
```
beczkowb, line 159: Do we need a separate test for running multiple workflows?

MarcinMoskala: No, but I separated it to have more smaller tests instead of one big

test/test_cli.py
MarcinMoskala: No, just different jobs and workflows. I check if when I run a workflow only this workflow is started, and when I start another one that only this other one is started.

PR #27, opened 2020-07-24
new deployment 

bartoszwalacik:


biggerquery/deploy.py
```diff
@@ -0,0 +1,111 @@
...
+from biggerquery.dagbuilder import get_dags_output_dir
+
+
+def os_call(cmd: List, input: str = None):
+    print('executing:', ' '.join(cmd))
+    if input:
+        run(cmd, check=True, input=input, encoding='ascii')
+    else:
+        run(cmd, check=True)
+    print('')
```
beczkowb, line 16: debugging leftover?

biggerquery/deploy.py
```diff
@@ -0,0 +1,111 @@
...
+    if auth_method == 'local_account':
+        return storage.Client(project=project_id)
+    elif auth_method == 'service_account':
+        oauthtoken = get_vault_token(deployment_config)
+        return storage.Client(project=project_id, credentials=credentials.Credentials(oauthtoken))
+    else:
+        raise ValueError('unsupported auth_method: ' + auth_method)
+
+
+def get_vault_token(deployment_config):
```
beczkowb, line 98: Do we have to use the Config class as a value object for these functions? 

Remember that Config won't be available during deployment (in a deployment plan the source is not available, parameters will be passed through the CLI). 

The second thing is that it's hard to say what the function needs (you need to read the function code). 

Thirdly, if we change something in the Config API, it will break the deployment.

biggerquery/deploy.py
```diff
@@ -0,0 +1,111 @@
...
+def os_call(cmd: List, input: str = None):
+    print('executing:', ' '.join(cmd))
+    if input:
+        run(cmd, check=True, input=input, encoding='ascii')
+    else:
+        run(cmd, check=True)
+    print('')
+
+
+def deploy_docker_image(build_ver: str, deployment_config, auth_method: str = 'local_account'):
```
iwko, line 19: Is it used anywhere?

test/test_deploy.py
```diff
@@ -0,0 +1,132 @@
...
+class DeployTestCase(TestCase):
+
+    def test_should_clear_GCS_DAGs_folder(self):
+
+        # given
+        workdir = os.path.dirname(__file__)
+        config = Config(name='dev',
+                        environment_variables_prefix='bamboo_bgq_',
+                        properties={
+       #                            'deploy_project_id': 'MY_DEPLOY_PROJECT_ID',
```
iwko, line 22: To remove?

test/test_deploy.py
```diff
@@ -0,0 +1,132 @@
...
+        monitoring_dag_blob.delete.assert_not_called()
+
+    def test_should_upload_files_from_DAGs_output_dir_to_GCS_DAGs_folder(self):
+
+        # given
+        workdir = os.path.dirname(__file__)
+        config = Config(name='dev',
+                        environment_variables_prefix='bamboo_bgq_',
+                        properties={
+                            # 'deploy_project_id': 'MY_DEPLOY_PROJECT_ID',
```
iwko, line 57: To remove?

biggerquery/deploy.py
```diff
@@ -0,0 +1,111 @@
+from google.oauth2 import credentials
```
iwko, line 1: All func args could have types

biggerquery/deploy.py
```diff
@@ -0,0 +1,111 @@
...
+
+def blob_URI(blob):
+    return f"gs://{blob.bucket.name}/{blob.name}"
+
+
+def upload_DAGs_folder(workdir: str, bucket):
+    dags_dir_path = get_dags_output_dir(workdir)
+
+    def upload_file(local_file_path, target_file_name):
+        print('bolb', target_file_name)
```
iwko, line 73: print('**blob**', target_file_name)

biggerquery/deploy.py
bartoszwalacik: nope, cmd output comes without eol

PR #34, opened 2020-08-07
Docs for Config

bartoszwalacik:


docs-configuration.md
```diff
@@ -0,0 +1,352 @@
...
+                           })
+```
+
+`Config` is the combination of Python `dicts` with
+some extra features that are useful for configuring things:
+
+1. The `Config` object holds multiple named configurations, one configuration
+per each environment (here `dev` and `prod`).
+
+1. Properties with constant values (here `vault_endpoint`)
```
iwko, line 36: Properties with constant values (here `vault_endpoint` and `docker_repository_project`)

docs-configuration.md
```diff
@@ -0,0 +1,352 @@
...
+```
+
+output:
+
+```text
+prod config:
+{'my_project': 'I_am_prod'}
+dev config:
+{'my_project': 'I_am_dev'}
+
```
iwko, line 230:
``` missing

File: docs-configuration.md [hunk 3]
```diff
@@ -0,0 +1,352 @@
...
+* `project_id` &mdash; GCP project Id of an internal dataset.
+* `dataset_name` &mdash; Internal dataset name.
+* `internal_tables` &mdash; List of table names in an internal dataset. 
+  Fully qualified names of internal tables are resolved to `{project_id}.{dataset_name}.{table_name}`.  
+* `external_tables` &mdash; Dict that defines aliases for external table names.
+  Fully qualified names of external tables have to be declared explicitly.
+
+The distinction between internal and external tables shouldn't be treated too seriously.
+Internal means `mine`. External means any other. It's just a naming convention.
```
iwko, line 316: Maybe: Internal means `mine` in your project?

docs-configuration.md
```diff
@@ -0,0 +1,352 @@
...
+You can change the roles by setting the `is_default` flag:
+
+```python
+config = Config(name='dev',
+                is_default=False,
+                properties={
+                   'my_project': 'I_am_{env}'
+                }).add_configuration(
+                name='prod',
+                is_default=True,
```
jkrasnod, line 215: what happens when:
- each configuration sets a flag value false?
- only first configuration sets a flag value false, second one doesn't specify it at all?

docs-configuration.md
```diff
@@ -0,0 +1,352 @@
...
+Having that, a `DatasetManager` instance can be easily created:
+
+```python
+dataset_manager = dataset_config.create_dataset_manager()
+``` 
+
+Then, you can use short table names in SQL, `DatasetManager` resolves them to fully qualified names: 
+
+```python
+dataset_manager.collect('select * from {quality_metric}')
```
jkrasnod, line 349: Maybe some examples of fully resolved queries (e.g. `select * from my-project-dev. scorer. quality_metric` on dev)?

docs-configuration.md
```diff
@@ -0,0 +1,352 @@
...
+dev config:
+{'my_project': 'I_am_dev'}
+```
+ 
+## Dataset Config
+
+`DatasetConfig` is a convenient extension to `Config` designed for workflows which use `DatasetManager`
+to call Big Query SQL. 
+
+`DatasetConfig` defines four properties that are required by `DatasetManager`:
```
jkrasnod, line 306: - Can I define some custom property and use it in a table name? 
- Can I use `{env}` in a table name?

bartoszwalacik: No, string interpolation works only for atomic String props

docs-configuration.md
bartoszwalacik: defaults are reasonable, by default in init method `is_default=True` and in add_configuration `is_default=False`. If you set all to False

PR #37, opened 2020-08-11
Rebrand

iwko:


bigflow/cli.py
```diff
@@ -9,12 +9,12 @@
 from types import ModuleType
 from typing import Tuple, Iterator
 import importlib.util
-import biggerquery as bgq
+import bigflow as bf
```
beczkowb, line 12: Maybe `bgf` instead of `bf`?

iwko: **b**ig**f**low => bf

PR #43, opened 2020-08-17
Cli final docs

bartoszwalacik:


docs/cli.md
```diff
@@ -189,25 +203,146 @@ BigFlow can build your workflows to Airflow DAGs and deploy them to Google Cloud
...
+Read more about `deployment_config.py` in
+[Managing configuration in deployment_config.py](#managing-configuration-in-deployment_configpy).
+
+
+### Building DAG files
+
+The `build-dags` command takes two optional parameters:
+
+* `--start-time` &mdash; the first [runtime](workflow-and-job.md#the-runtime-parameter)
+  of your workflows. If empty, a `current hour` is used for hourly workflows and
```
beczkowb, line 238: Why  `current hour` and `today` are written as code?

beczkowb:
In general - approved

docs/cli.md
bartoszwalacik: it's not now, minutes and seconds are truncated

bartoszwalacik: `--start-time` &mdash; the first [runtime](workflow-and-job.md#the-runtime-parameter)
  of your workflows. If empty, a current hour (`datetime.datetime.now().replace(minute=0, second=0, microsecond=0)`)
  is used for hourly workflows and `datetime.date.today()` for daily workflows.

PR #48, opened 2020-08-24
vault fetch debugging, default_project_setup, find_or_create_setup_fo…

beczkowb:
…r_main_project_package

improved versioning
adding Workflow.start_time_expression_factory, removing Workflow.runtime_as_datetime

### DIFF
--- a/bigflow/build.py
+++ b/bigflow/build.py
@@ -134,6 +134,7 @@ def build_command(
             ('start-time=', None, 'DAGs start time -- given in local timezone, for example: 2020-06-27 15:00:00'),
             ('workflow=', None, 'The workflow that you want to build DAG for.'),
             ('export-image-to-file', None, 'If used, builder will save the Docker image to tar and remove image from the local registry.'),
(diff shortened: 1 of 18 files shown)


bigflow/version.py
```diff
@@ -1,46 +1,64 @@
 import re
+from typing import Optional
+from uuid import uuid1
```
iwko, line 3: why uuid1?

beczkowb: https://docs.python.org/3.7/library/uuid.html#uuid.uuid1
looks like the best option

bigflow/build.py
```diff
@@ -112,6 +113,12 @@ def export_image_to_file(tag, image_dir, version):
     remove_docker_image_from_local_registry(tag)
 
 
+class MetaCommand(type):
+    # For test_build
```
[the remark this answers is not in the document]
beczkowb: done

iwko (approved):

PR #51, opened 2020-08-27
Dependencies cleaning

iwko:


bigflow/clear_methods.py
```diff
@@ -0,0 +1,18 @@
+from pathlib import Path
```
beczkowb, line 1: I think 18 lines of code, used only in 1 module, is not enough to make a new module

requirements/bigquery_extras.txt
```diff
@@ -0,0 +1,5 @@
+google-cloud-core==1.4.1
+google-cloud-storage==1.30.0
```
beczkowb, line 2: Do we need cloud storage?

requirements/bigquery_extras.txt
```diff
@@ -0,0 +1,5 @@
+google-cloud-core==1.4.1
```
beczkowb, line 1: Do we need cloud core?

bigflow/clear_methods.py
beczkowb: Sorry, now I see that you use it in another modules, but still I think it should be in the build module, because it clears build leftovers.

bigflow/__init__.py
```diff
@@ -1,19 +1,24 @@
...
 from .configuration import Config
-from .build import default_project_setup
+
 
 __all__ = [
     # core
     'Workflow',
     'Definition',
     'Config',
-    'default_project_setup',
```
beczkowb, line 13: why?

bigflow/bigquery/__init__.py
```diff
@@ -1,13 +1,17 @@
...
+    __all__ = [
+        'component',
+        'add_label_component',
+        'sensor_component',
+        'INLINE_COMPONENT_DATASET_ALIAS',
+        'DatasetConfig',
+        'Job'
+    ]
+except ImportError:
+    pass
```
beczkowb, line 16: In this case `__all__` should be empty I think.

setup.py
```diff
@@ -48,15 +51,16 @@ def run(self):
     url="https://github.com/allegro/bigflow",
     packages=setuptools.find_packages(exclude=('test', 'e2e')),
     data_files=[
-        ('requirements', ['requirements/base.txt', 'requirements/stackdriver_extras.txt']),
+        ('requirements', ['requirements/base.txt', 'requirements/stackdriver_extras.txt', 'requirements/bigquery_extras.txt']),
```
beczkowb, line 54: Lets change "stackdriver" to more generic "monitoring" (in the whole project).

bigflow/clear_methods.py
iwko: It cannot be in build modules because build is imported in setup.py and setup.py imports xmlrunner. At the moment of using setup.py xmlrunner is not installed yet so we get ImportError

requirements/bigquery_extras.txt
iwko: Yes

bigflow/bigquery/__init__.py
iwko: and it is :)

requirements/bigquery_extras.txt
iwko: yes

bigflow/__init__.py
iwko: It is not needed, and also it results in importing build module and trying to access xmlrunner

bigflow/bigquery/__init__.py
iwko: Done

setup.py
iwko: Done

PR #74, opened 2020-09-29
 Change Workflow arg start_time_{expression => ''}_factory

anjensan:
Change 'bigflow.Workflow' constructor argument start_time_expression_factory:
 - renamed to `start_time_factory`
 - type of callback changed from str->str to datetime->datetime
 - dagbuilder.* functions accept both 'str' (old variant) and datetime objects
 - generated dags contains 'repr(<`datetime obj>)`
 - documentation is *not* updated (is under work in another PR).

### DIFF
--- a/bigflow/dagbuilder.py
+++ b/bigflow/dagbuilder.py
@@ -1,4 +1,6 @@
 import shutil
+import typing
+
 from pathlib import Path
 from datetime import datetime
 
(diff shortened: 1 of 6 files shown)


bigflow/dagbuilder.py
```diff
@@ -109,4 +112,13 @@ def get_dags_output_dir(workdir: str) -> Path:
     if not dags_dir_path.exists():
         dags_dir_path.mkdir()
 
-    return dags_dir_path
\ No newline at end of file
+    return dags_dir_path
+
+
+def _coerce_str_to_datetime(dt: typing.Union[str, datetime]):
```
beczkowb, line 118: Maybe just `str_to_datetime`?

Instead of using `_`, add non-private module elements to `__all__`.

anjensan: Function accepts both str & datetime, this is why it prefixed with `_coerce`
But I really don't have strong preference, because it is private: so it may be '_as_datetime', 'str_to_datetime' etc. Changed.

I'm not sure what functions should be in '__all__' (all existing?)
Also not all people check existing of '__all__' (for example I don't ;) and still may use some 'private/hidden' functions from module.
So I kept underscore for now, but can remove it & add '__all__' if you want.

beczkowb:
approved

PR #84, opened 2020-10-08
Update Stackdriver logging

anjensan:
Various updates in logging:
 - Add config option 'log_config' to `Workflow` class
 - Automatically configure logging when 'bigflow run' command is executed (via Workflow.init_logging)
   User still need to invoke `workflow.init_logging` when workflow is used from custom script (as many various workflows may be mixed together).
 - Change logging configuration API: rename 'bigflow.log.BigFlowLogging.configure_logging(..)' => 'bigflow.log.init_logging' & change its args (consumes config dict); 'workflow_id' is mandatory, 'log_name' used only as Stackdriver log.
 - Refactor: remove unused methods from log Handler (info, warning etc)
 - Change attached labels: 'id' => 'workflow_id', add random unique uuid (may be used to filter logs of one single workflow execution).
 - Use structured (json) logging instead of text: also include module name, filename, lineno, stacktrace etc.
 - Update documentation.


bigflow/cli.py
```diff
@@ -127,7 +129,9 @@ def execute_workflow(root_package: Path, workflow_id: str, runtime=None):
 
     @param runtime: str determine partition that will be used for write operations.
     """
-    find_workflow(root_package, workflow_id).run(runtime)
+    w = find_workflow(root_package, workflow_id)
```
beczkowb, line 132: I don't like the idea that workflow initializes logging. It already has 3 responsibilities:

* running jobs
* building execution order
* holding DAG configuration

I think you can move the logging initialization to a function, and the Workflow class can act only as logging_configuration (and probably other kinds of configuration) holder.

What do you think?

bigflow/scaffold/scaffold_templates.py
```diff
@@ -48,12 +48,14 @@
...
-                               'dags_bucket': '{dags_bucket}'}})
+advanced_deployment_config_template = '''.add_configuration(
+                            name='{env}',
+                            properties={{
+                                'gcp_project_id': '{project_id}',
+                                'dags_bucket': '{dags_bucket}',
+                            }})
 '''
-requirements_template = f'''bigflow[bigquery]==1.0.dev33
+requirements_template = '''bigflow[bigquery,log]=={bigflow_version}
```
beczkowb, line 58: Good idea 👍

docs/logging.md
```diff
@@ -1,58 +1,68 @@
...
 
-```python
-from bigflow.log import BigflowLogging
-project_id = 'some_project'
-logger_name = __name__
-
-
-BigflowLogging.configure_logging(project_id, logger_name)
-```
+To use logging, you have to install `bigflow` with `log` module - `bigflow[log]`.
```
beczkowb, line 7: To use logging, you have to install `bigflow` with `log` module:

`pip install bigflow[log]`

bigflow/cli.py
iwko: I don't like initialization in the Workflow. Also user can install bigflow without our `log` module so this code will throw Exception

bigflow/log.py
```diff
@@ -1,89 +1,136 @@
...
+                    "project_id": str(self.project_id),
+                },
+            },
+        )
+
+    def create_logging_client(self):
+        return logging_v2.LoggingServiceV2Client()
+
+    def emit(self, record: logging.LogRecord):
+        cl_log_level = record.levelname  # CloudLogging list of supported log levels is a superset of python logging level names
```
iwko, line 44: I think we can remove this comment

bigflow/__init__.py
```diff
@@ -2,6 +2,8 @@
 from .workflow import Workflow, Definition
 from .configuration import Config
 
+from bigflow._version import __version__
```
beczkowb, line 5: Do we need a separate module to hold the version?

bigflow/log.py
```diff
@@ -1,89 +1,136 @@
...
 
-def excepthook(logger):
-    sys.excepthook = handle_uncaught_exception(logger)
\ No newline at end of file
+    logging.info(dedent(f"""
+           *************************LOGS LINK*************************
+           Workflow logs (all runs): {workflow_logs_link}
+           Only this run logs: {this_execution_logs_link}
+           ***********************************************************"""))
+    logging.getLogger(None).addHandler(gcp_logger_handler)
```
iwko, line 135: logging.getLogger().addHandler(gcp_logger_handler)

bigflow/log.py
```diff
@@ -1,89 +1,136 @@
...
 
+try:
+    from typing import TypedDict
+except ImportError:
+    TypedDict = dict
 
-def create_logging_client():
-    return logging_v2.LoggingServiceV2Client()
 
+class GCPLoggerHandler(logging.Handler):
```
beczkowb, line 19: Why do you inherit from the Handler class if you use the StreamHandler Init?

docs/logging.md
```diff
@@ -1,58 +1,68 @@
...
+    name='workflow_name',
+    log_config={
+        'project_id': gcp_project_id,
+        'level': 'DEBUG',
+    },
+    # ...
+)
+```
 
+BigFlow will automatically configure `logging` when workflow is executed with `bigflow run` or `bigflow deploy` commands.
```
iwko, line 25: I don't like `automatically` idea. Maybe not every user want logs in CL.

bigflow/log.py
```diff
@@ -1,89 +1,136 @@
...
     return handler
 
 
-class BigflowLogging(object):
-    IS_LOGGING_SET = False
+_LOGGING_CONFIGURED = False
+
+
+# TODO: move to some shared config module?
+class LogConfigDict(TypedDict):
```
beczkowb, line 86: +1

test/test_excepthook.py
```diff
@@ -1,12 +0,0 @@
-import logging
```
iwko, line 1: Why did you remove this test?

bigflow/__init__.py
iwko: Imo it's clear where you can find current version

docs/logging.md
beczkowb: Logging is not configured out of the box (unfortunately, we don't have the project_id to do that). 

Btw. default, reasonable configuration is a good idea in my opinion.

bigflow/log.py
```diff
@@ -1,89 +1,136 @@
...
+            'message': record.getMessage(),
+            'name': record.name,
+            'filename': record.filename,
+            'module': record.module,
+            'lineno': record.lineno,
         }
 
-    def warning(self, message):
-        self.write_log_entries(message, 'WARNING')
+        if record.exc_info and not record.exc_text:
```
iwko, line 53: It's not clear to me what's going on in lines from 53-59.

bigflow/log.py
```diff
@@ -1,89 +1,136 @@
...
-        labels = {"id": f"{self.workflow_id or self.project_id}"}
-        entry = logging_v2.types.LogEntry(
-            log_name=log_name,
-            resource=self.get_resource(),
-            text_payload=message,
-            severity=severity,
-            labels=labels
-        )
-        self.client.write_log_entries([entry])
+    def _format_exception(self, ei):
```
iwko, line 64: what's ei?

bigflow/log.py
```diff
@@ -1,89 +1,136 @@
...
+
+def _generate_cl_log_view_link(params: dict):
+    query = quote_plus("\n".join(
+        '{}="{}"'.format(k, v)
+        for k, v in params.items()
+    ))
+    return f"https://console.cloud.google.com/logs/query;query={query}"
+
+ 
+def init_logging(config: LogConfigDict, workflow_id=None, force=False):
```
iwko, line 100: What's the force parameter? Is it used anywhere?

bigflow/log.py
```diff
@@ -1,89 +1,136 @@
...
+    global _LOGGING_CONFIGURED
+    if _LOGGING_CONFIGURED and not force:
+        import warnings
+        warnings.warn(UserWarning("bigflow.log is is already configured - skip"))
+        return
+
+    _LOGGING_CONFIGURED = True
+
+    gcp_project_id = config['gcp_project_id']
+    workflow_id = workflow_id or config.get('workflow_id', "unknown-workflow")
```
iwko, line 111: I dont think we should use unknown-workflow. IMO we should throw if worklfow_id or config.workflow_id  not provided

bigflow/log.py
```diff
@@ -1,89 +1,136 @@
...
+    log_name = config.get('log_name', workflow_id)
+    log_level = config.get('log_level', 'INFO')
+    run_uuid = str(uuid.uuid4())
```
iwko, line 114: 👍

bigflow/workflow.py
```diff
@@ -23,24 +30,52 @@ def daily_start_time(start_time: dt.datetime) -> dt.datetime:
...
+            return
+        log.init_logging(
+            self.log_config,
+            workflow_id=self.workflow_id,
+        )
+
+    def _check_logging_initialized(self):
+        if self.log_config and not self._log_initialized:
+            import warnings
+            warnings.warn(UserWarning("Log configuration is attached to the Workflow, but method `.init_logging() is not called"))
```
iwko, line 68: So maybe if config provided init_logging() should be called in __init__

docs/logging.md
```diff
@@ -1,58 +1,68 @@
 # Logging
 
-Using Bigflow, you can easily add a handler to your root logger that will send all logging calls to the Cloud Logging.
-In the CL, you can filter logs per project id/logger name and workflow id. 
 
-## Getting started
-To use logging, you have to install `bigflow` with `logger` module - `bigflow[log]`.
-To create a logger with the CL handler, you have to use `configure_logging` method.
+With Bigflow you can easily configure Python logging to send all messages the Cloud Logging (Stackdriver).
```
iwko, line 4: what messages? maybe logs?

docs/logging.md
```diff
@@ -1,58 +1,68 @@
...
+Cloud Logging allows you to filter logs by project id / log name / workflow id and other fields.
```
iwko, line 5: what other fields?

docs/logging.md
```diff
@@ -1,58 +1,68 @@
...
+    workflow.run("2018-01-01")
+    workflow.run("2018-01-02")
 ```
-After calling the `configure_logging` method, your calls to any logger will be send to CL.
 
-Also in a console, you should see a
-link to the CL query. In the case of the code above, the link should look like this `https://console.cloud.google.com/logs/query;query=logName%3D%22projects%2Fsome-project%2Flogs%some-logger%22%0Alabels.id%3D%some-workflow%22\`
-
-To send logs to the CL, you use the standard python `logging` module.
+In order to send logs into the Cloud Logging you should the standard python `logging` module.
```
iwko, line 48: In order to send logs into the Cloud Logging you should use the standard python `logging` module.

setup.py
```diff
@@ -19,9 +20,15 @@
     log_extras_require = [l.strip() for l in log_extras_requirements.readlines()]
 
 
+with open(os.path.join('bigflow', '_version.py'), 'r') as version_file:
```
iwko, line 23: Why can't we just import bigflow._version?

test/test_log.py
```diff
@@ -1,150 +1,115 @@
...
+        self.assertEqual(le.log_name, "projects/project-id/logs/logger_name")
+        
+        if severity:
+            self.assertEqual(le.severity, severity)
+        if message_re:
+            self.assertRegexpMatches(le.json_payload['message'], message_re)
+
+        return le
 
     def test_should_log_unhandled_exception(self):
```
iwko, line 72: Imo previous test was better, but this one is simpler.

bigflow/__init__.py
anjensan: A separate file. It is 'exec'ed from setup.py directly (so bigflow/__init__.py is not involved).
We can't put version to '__init__' directly, because otherwise 'setup.py' will try to import 'bigflow' before it is installed (or we need to duplicate version in 2 places).

bigflow/log.py
iwko: Also LogConfigDict doesnt have workflow_id property

bigflow/log.py
anjensan: exc_info, changed.

bigflow/__init__.py
beczkowb: true

setup.py
anjensan: Because it will trigger loading of '__init__.py', which will try to import the whole BigFlow.

iwko: So maybe it will be simpler to hold VERSION in setup.py?

docs/logging.md
anjensan: Technically StackDriver allows you to filter by any available field.
In our case it might be severity, timestamp, python module/filename (with structured logging) etc.
Not sure that we need to list all of them here - CL provides UI for this (click on the field).

setup.py
beczkowb: Then it's not available in the bigflow package.

bigflow/log.py
anjensan: exc_info, changed

setup.py
beczkowb: I vote for the Andrei solution.

test/test_excepthook.py
anjensan: We need to verify that 'exchook' is installed, but not that it is actually called (because it is responsibility of Python itself). So there is no need to spawn external process etc, I've simplified that test.

iwko: Ok, 👍

docs/logging.md
iwko: Maybe provide some link here to stackdriver docs instead of writing `another fields`

setup.py
iwko: 👍

bigflow/cli.py
beczkowb: On the other hand, if you want to run a workflow without the CLI, and still have the logging initialized, it's problematic...

anjensan: In fact it is already in a separate function, Workflow just expose wrapper for it.
Agree, that it is better not to add such logic to Workflow.
Changed: Dropped `Workflow.init_logging()`, expose attribute `Workflow.log_config`.

bigflow/log.py
anjensan: I'm going to do it later.

bigflow/log.py
anjensan: Hm, I'd like to keep it (at least for me it is not obvious that we don't need to perform sort of mapping between python & CL logging levels).

bigflow/log.py
anjensan: ack

bigflow/log.py
anjensan: This parameter overrides behaviour of _LOGGING_CONFIGURED.
IMO it is not needed at this time, removed.

bigflow/log.py
anjensan: Agree. 'workflow_id' become required parameter.

bigflow/workflow.py
anjensan: IMO it is not a best place for it.

docs/logging.md
anjensan: May be 'log messages'. done

docs/logging.md
anjensan: done.

test/test_log.py
anjensan: So, should I apply any changes to test?

bigflow/log.py
beczkowb: I wanted to say that I like the TypedDict :)

docs/logging.md
iwko: 👍

bigflow/workflow.py
iwko: Why

bigflow/log.py
anjensan: added comment

docs/logging.md
anjensan: I've removed 'other fields' completely from documentation, as it should be small and concise.

bigflow/workflow.py
anjensan: Because:
- It might be hard to test code, where __init__ contains some side effects (modification of env, logging etc)
- bigflow command tries to walk across all python modules, which triggers creation of all workflows in project - so all Workflow.__init__ will be called multiple times "by-design"
- actually putting this into __ init __ will mean "initialization in the Workflow", which was discussed in other comment-thread.

iwko (approved):

PR #88, opened 2020-10-13
Fix log configuration

anjensan:
- Update docs
- Some libs (like apache beam) may configure logging by themselves (with logging.basicConfig)
   Clear existing logging when 'bigflow.log.init_logging' is used.

### DIFF
--- a/bigflow/log.py
+++ b/bigflow/log.py
@@ -16,6 +16,9 @@ except ImportError:
     TypedDict = dict
 
 
+logger = logging.getLogger(__name__)
+
+
 class GCPLoggerHandler(logging.Handler):
 
     def __init__(self, project_id, log_name, labels):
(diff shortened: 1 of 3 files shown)


bigflow/log.py
```diff
@@ -127,7 +130,13 @@ def init_logging(config: LogConfigDict, workflow_id: str):
         'run_uuid': run_uuid,
     }
 
+    # clean all existings handlers
+    root = logging.getLogger(None)
```
beczkowb, line 134: I'm not sure if we want to clean the existing configuration. For example, Dataflow provides nice logging UI utilities and we want to keep them. 

I suggest making BigFlow logging as an extension (so BigFlow logging configuration sends logs to an additional place).

test/test_log.py
```diff
@@ -27,18 +28,25 @@ def setUp(self):
...
     def tearDown(self):
-        logging.getLogger().handlers.clear()
+        self._clear_all_root_loggers()
         bigflow.log._LOGGING_CONFIGURED = False
 
     def test_should_create_correct_logging_link(self):
-        with self.assertLogs(level='INFO') as logs:
+
+        f = io.StringIO()
+        with contextlib.redirect_stderr(f):
```
iwko, line 44: Why stderr?

anjensan: Because by default 'logging.basicConfigure()' just creates 'logging.StreamHandler(None)' and attaches it to the root logger.  StreamHandler output is 'sys.stderr' by default.

bigflow/log.py
anjensan: It makes sense.
I've changed logic to do 2 things:
 - install 'stderr' logger only when no other loggers are installed (otherwise most probably either console logger is already installed OR is not needed at all)
 - update root log-level if it greater than specified log_level.

beczkowb:
approved

PR #90, opened 2020-10-15
Extend Job api

anjensan:
Extend Job api.

When job has method 'execute', it is executed with 'workflow' object, runtime (parsed as datetime), workflow id
Otherwise the old method 'run(runtime)' is executed.  This change does not break all existing jobs.
Eventually we may deprecate/remove method 'run' from jobs.

### DIFF
--- a/bigflow/workflow.py
+++ b/bigflow/workflow.py
@@ -43,17 +43,14 @@ class JobContext(typing.NamedTuple):
 class Job(abc.ABC):
     """Base abstract class for bigflow.Jobs.  It is recommended to inherit all your jobs from this class."""
 
-    @abc.abstractproperty
-    def get(self):
+    @property
+    @abc.abstractmethod
+    def id(self) -> str:
         raise NotImplementedError
 
     @abc.abstractmethod
-    def run_job(self, context: JobContext):
-        logger.warn("Call old-api method 'job.run()' for %s", self.id)
-        self.run(context.runtime_raw)
-
-    def run(self, runtime):
-        raise NotImplementedError("Methods 'run_job' is not implemented for %r" % self)
+    def execute(self, context: JobContext):
+        raise NotImplementedError
 
 
 class Workflow(object):
@@ -85,11 +82,11 @@ class Workflow(object):
                 pass
         raise ValueError("Unable to parse 'run_time' %r" % rt)
 
-    def _apply_run_job(self, job, context):
+    def _execute_job(self, job, context):
         if not isinstance(job, Job):
             logger.warn("Please, inherit your job %r from `bigflow.Job` class", job)
-        if hasattr(job, 'run_job'):
-            job.run_job(context)
+        if hasattr(job, 'execute'):
+            job.execute(context)
         else:
             # fallback to old api
             job.run(context.runtime_raw)
@@ -113,7 +110,7 @@ class Workflow(object):
     def run(self, runtime: typing.Union[dt.datetime, str, None] = None):
         context = self._make_job_context(runtime)
         for job in self.build_sequential_order():
-            self._apply_run_job(job, context)
+            self._execute_job(job, context)
 
     def find_job(self, job_id):
         for job_wrapper in self.build_sequential_order():
@@ -123,7 +120,7 @@ class Workflow(object):
 
     def run_job(self, job_id: str, runtime: typing.Union[dt.datetime, str, None] = None):
         context = self._make_job_context(runtime)
-        self._apply_run_job(self.find_job(job_id), context)
+        self._execute_job(self.find_job(job_id), context)
 
     def build_sequential_order(self):
         return self.definition.sequential_order()
--- a/bigflow/workflow.py
+++ b/bigflow/workflow.py
@@ -1,19 +1,21 @@
 import abc
-import datetime as dt
+import collections
 import typing
-
-from collections import OrderedDict
-from typing import Optional
+import warnings
+import datetime as dt
 
 import bigflow
-from bigflow.commons import now
-
 
 import logging
 logger = logging.getLogger(__name__)
 
 
 DEFAULT_SCHEDULE_INTERVAL = '@daily'
+
+_RUNTIME_FORMATS = [
+    "%Y-%m-%d %H:%M:%S",
+    "%Y-%m-%d",
+]
 
 
 def get_timezone_offset_seconds() -> int:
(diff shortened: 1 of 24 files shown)


bigflow/workflow.py
```diff
@@ -29,8 +30,36 @@ def daily_start_time(start_time: dt.datetime) -> dt.datetime:
     return start_time.replace(hour=0, minute=0, second=0, microsecond=0) - td
 
 
+class JobContext(typing.NamedTuple):
+    runtime_raw: typing.Optional[str]
+    runtime: typing.Optional[dt.datetime]
+    env: typing.Optional[str]
```
beczkowb, line 36: Do we really need `env` in a job context? Up to this point, env was only known for the `Configuration` class, and the CLI, which only sets an environmental variable.

iwko:
CI is failling

bigflow/workflow.py
anjensan: No, it was committed by mistake. Removed.

bigflow/workflow.py
```diff
@@ -29,8 +30,36 @@ def daily_start_time(start_time: dt.datetime) -> dt.datetime:
...
+class Job(abc.ABC):
+    """Base abstract class for bigflow.Jobs.  It is recommended to inherit all your jobs from this class."""
+
+    @property
+    @abc.abstractmethod
+    def id(self) -> str:
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def execute(self, context: JobContext):
```
beczkowb, line 50: I'm not sure if we want to change the interface so drastically. 

* Most of the jobs won't use all of these parameters, only the runtime parameter.
* We have many jobs to change in our analytics, even though they don't need the additional context.
* It sets the bar a bit higher at the beginning (single parameter `runtime` is easier to understand than the rich `JobContext`).

I suggest something in the middle:
```
@abc.abstractmethod
def run(self, runtime, additional_context: Optional[JobContext]):
    pass
```

In that case, if a user doesn't need additional context, a job can be very simple:
```
class MyJob:
    id = 'my_job'
    
    def run(self, runtime):
        pass
```

And if a user needs some additional context info, then takes an additional argument:
```
class MyJob:
    id = 'my_job'
    
    def run(self, runtime, additional_context):
        pass
```

bigflow/workflow.py
```diff
@@ -45,24 +74,53 @@ def __init__(
...
+        for fmt in self.RUNTIME_FORMATS:
+            try:
+                return dt.datetime.strptime(rt, fmt)
+            except ValueError:
+                pass
+        raise ValueError("Unable to parse 'run_time' %r" % rt)
+
+    def _execute_job(self, job, context):
+        if not isinstance(job, Job):
+            logger.warn("Please, inherit your job %r from `bigflow.Job` class", job)
```
beczkowb, line 85: Currently, a user has no need to inherit from the interface, so why should he? It's an additional thing to:

* know about
* remember
* do

bigflow/workflow.py
```diff
@@ -45,24 +74,53 @@ def __init__(
         self.start_time_factory = start_time_factory
         self.log_config = log_config
 
-    def run(self, runtime: Optional[str] = None):
-        if runtime is None:
-            runtime = self._auto_runtime()
+    def _parse_runtime_str(self, rt: str):
```
beczkowb, line 75: `rt` -> `runtime`, `fmt` -> `format`

bigflow/workflow.py
```diff
@@ -29,8 +30,36 @@ def daily_start_time(start_time: dt.datetime) -> dt.datetime:
...
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def execute(self, context: JobContext):
+        raise NotImplementedError
+
+
 class Workflow(object):
 
+    RUNTIME_FORMATS = [
```
beczkowb, line 56: Why do we need to support multiple runtime formats? Currently, for a deployed workflow, a user can't choose the format of a passed runtime.

anjensan: AFAIK in most places we use '%y-%m-%d' as runtime format.
But also when runtime is not specified - it is autogenerated to be in '%y-%m-%d %H:%M:%S' format.

bigflow/workflow.py
anjensan: done.

bigflow/workflow.py
anjensan: I see two reasons:
 - it might be easier to add further changes/features to 'bigflow' without breaking existing code (so it more like a 'good-practice');
 - better support from ide/editors
    for example in my case i type "def exe<ENTER>" and editor (vscode) complete it to "def execute(self, context: JobContext):" automatically; and I have autocompletion for 'context.' variable.

bigflow/workflow.py
anjensan: Hm, I thought about this option. 
However it requires slightly more runtime magic - inspecting method signature (checking number of args).
Python doesn't have overloading by number of arguments, so it looks like attempt to use idiom from other languages.
IMO It makes harder to understand the code for external contributors and users.

bigflow/workflow.py
beczkowb: Users can run a workflow with runtime formatted in any way. But when scheduled, the only available format is date-time:

dagbuilder.py
```
 arguments=['run', '--job', '{bf_job}', '--runtime', '{{{{ execution_date.strftime("%Y-%m-%d %H:%M:%S") }}}}', '--project-package', '{root_folder}', '--config', '{{{{var.value.env}}}}'],
```

Handling different formats is currently strictly related to the CLI run. And as I can see, CLI already has that logic:

```
def _valid_datetime(dt):
    try:
        datetime.strptime(dt, "%Y-%m-%d %H:%M:%S")
        return dt
    except ValueError:
        try:
            datetime.strptime(dt, "%Y-%m-%d")
            return dt
        except ValueError:
            raise ValueError("Not a valid date: '{0}'.".format(dt))
```

anjensan: I saw '%y-%m-%d' in unit tests. https://github.com/allegro/bigflow/blob/master/test/test_workflow.py#L17
https://github.com/allegro/bigflow/blob/master/test/test_scaffold_bigquery_example.py#L23
I thought it is an API, do we need to update tests?

As I can see _validate_datetime only performs validation, it doesn't change actual 'dt' string.

bigflow/workflow.py
anjensan: Also the idea was to add parsed version of 'runtime', so user don't need to deal with parsing, different format (and care about this format at all).  That is why 'context' include 'runtime' and 'runtime_raw'.

bigflow/workflow.py
beczkowb: "As I can see _validate_datetime only performs validation, it doesn't change actual 'dt' string." - true. I think it can be useful for daily workflows (so a user doesn't have to pass time every time running `bf run`).

"I saw '%y-%m-%d' in unit tests. https://github.com/allegro/bigflow/blob/master/test/test_workflow.py#L17
https://github.com/allegro/bigflow/blob/master/test/test_scaffold_bigquery_example.py#L23
I thought it is an API, do we need to update tests?" - There should be a test for each format in this case.

bigflow/workflow.py
beczkowb: **better support from ide/editors
for example in my case i type "def exe" and editor (vscode) complete it to "def execute(self, context: JobContext):" automatically; and I have autocompletion for 'context.' variable.**

That's a good point. I'm not against that this interface exists, I just don't like the idea of forcing users to use it without a clear reason. If some want to use it for IDE - just do it.

**it might be easier to add further changes/features to 'bigflow' without breaking existing code (so it more like a 'good-practice');**

* if we really want to force users to use this interface, for any reason, then warn is not enough. 
* If we decide to add some features related to the interface, most users still get breaking changes.
* Python has duck-typing, we should depend on the behavior and avoid type checking.
* "we might need it in the future" we don't know, but we know it has a cost right now.

I leave that for @iwko @bartoszwalacik to decide.

bigflow/workflow.py
beczkowb: Well, that's true that doing Java from Python is not the best idea. The second argument is also true (though these operations are super simple, and we need to maintain, test, and document them).

I suggest that in the first version we keep the execute method out of the public interface, for the internal utils like dataproc job etc. **Making it public and recommended requires a lot of work in the documentation and our projects**. What is more, we don't need it in our codebase at this moment.

@bartoszwalacik @iwko what do you think?

beczkowb (commented):
reviewed

bigflow/workflow.py
```diff
@@ -29,8 +30,34 @@ def daily_start_time(start_time: dt.datetime) -> dt.datetime:
     return start_time.replace(hour=0, minute=0, second=0, microsecond=0) - td
 
 
+class JobContext(typing.NamedTuple):
+    runtime: typing.Optional[dt.datetime]
+    runtime_raw: typing.Optional[str]
```
bartoszwalacik, line 35: runtime_as_str

bigflow/workflow.py
```diff
@@ -29,8 +30,34 @@ def daily_start_time(start_time: dt.datetime) -> dt.datetime:
...
+    workflow: typing.Optional['Workflow']
+    workflow_id: typing.Optional[str]
```
bartoszwalacik, line 37: why not workflow.workflow_id ?

anjensan: I was assuming (incorrectly) hat it might be a situation when we don't have a reference to the Workflow, but know it's id.
I've removed 'workflow_id' from JobContext

docs/workflow-and-job.md
```diff
@@ -45,26 +47,31 @@ Running workflows and jobs from a module is useful for debugging. In any other c
...
 ## Job
 
-A job is just an object with a unique `id` and the `run` method.
+A job is just an object with a unique `id` and the `execute` method.
 
 The `id` parameter is a string that should be a valid Python variable name. For example — `'my_example_job'`, `'MY_EXAMPLE_JOB'`, `'job1234'`.
 
-The Job `run` method a single argument — `runtime`. The `runtime` parameter is a data-time string. 
-You can find more information about `runtime` and scheduling [workflow scheduling options](#workflow-scheduling-options).
+The Job `run` method a single argument — `context`. The `context` parameter is an object of type `bigflow.JobContext`.
```
beczkowb, line 54: `run` -> `execute`

beczkowb:
To be clear:

- You add the new job execution API
- The old API is only for backward compatibility in our projects
- You don't mention the old API in docs
- You will update the remaining docs in another PR

@anjensan am I right?

If yes, approved.

anjensan:
@beczkowb 
yes :)

PR #92, opened 2020-10-15
Dataproc - integration with PySpark

anjensan:
This PR contains implementation/tests for PySparkJob
Does not include documentation (wip) / scaffolding (wip).


bigflow/dataproc.py
```diff
@@ -0,0 +1,358 @@
...
+
+
+logger = logging.getLogger(__name__)
+
+__all__ = [
+    'PySparkJob',
+]
+
+DEFAULT_REQUIREMENTS = [
+    'bigflow[dataproc,log,bigquery]==1.0.dev67',
```
beczkowb, line 34: Get the version from the `_version` module.

bigflow/dataproc.py
```diff
@@ -0,0 +1,358 @@
...
+        bucket_id: str,
+        gcp_project_id: str,
+        gcp_region: str,
+        driver_arguments: typing.Optional[dict] = None,
+        pip_packages: typing.Union[typing.Iterable[str], pathlib.Path] = (),
+        jar_file_uris: typing.Iterable[str] = ('gs://spark-lib/bigquery/spark-bigquery-latest_2.12.jar',),
+        worker_num_instances: int = 2,
+        worker_machine_type: str = 'n1-standard-1',
+        env: str = '',
+        setup_file: str = '',
```
beczkowb, line 53: Why an empty string instead of None?

bigflow/dataproc.py
```diff
@@ -0,0 +1,358 @@
...
+            self.setup_file = setup_file
+        else:
+            self.setup_file = None
+            self._project = _capture_caller_topmodule()
+            self._any_file_inside_project = _capture_caller_path(1)
+
+    def _ensure_has_setup_file(self):
+        if self.setup_file:
+            return
+        logger.debug("Find or create setup file for main package")
```
beczkowb, line 88: Can we use only 3 levels of logging - `info`, `warn`, `error`? It's the system that most of the people in Allegro use and works it works fine.

bigflow/configuration.py
```diff
@@ -10,12 +11,9 @@ def __init__(self,
...
 
 
-class StringStream:
-    def __init__(self):
-        self.value = ''
-
-    def write(self, text):
-        self.value += text
+def current_env():
+    """Returns current env name (specified via 'bighflow --config' option)"""
```
beczkowb, line 15: bighflow -> bigflow

bigflow/configuration.py
```diff
@@ -10,12 +11,9 @@ def __init__(self,
...
+    return os.environ.get('env')
```
beczkowb, line 16: I think it's `bf_env`:

```
def set_configuration_env(env):
    """
    Sets 'bf_env' env variable
    """
    if env is not None:
        os.environ['bf_env'] = env
        print(f"bf_env is : {os.environ.get('bf_env', None)}")
```

BTW. the `bf_` part is the default value for `environment_variables_prefix` in `Config`. The problem is that CLI always assumes that the `environment_variables_prefix` value is `bf_`.

bigflow/dataproc.py
anjensan: Agree, better to use Optional[str] = None, fixed.

bigflow/configuration.py
anjensan: Fixed.

beczkowb:
@anjensan approved

PR #94, opened 2020-10-19
Beam job util

beczkowb:
Adding `BeamJob` which takes pipeline options and driver, then enriches the pipeline options and runs the job.

### DIFF
--- a/bigflow/__init__.py
+++ b/bigflow/__init__.py
@@ -26,9 +26,3 @@ try:
     __all__.append('monitoring')
 except ImportError:
     pass
-
-try:
-    from . import dataflow
-    __all__.append('dataflow')
-except ImportError:
-    pass
--- a/bigflow/__init__.py
+++ b/bigflow/__init__.py
@@ -12,7 +12,7 @@ __all__ = [
     'JobContext',
     'Definition',
     'Config',
-    'resources',
+    'resources'
 ]
 
 try:
--- a/bigflow/__init__.py
+++ b/bigflow/__init__.py
@@ -26,3 +26,9 @@ try:
     __all__.append('monitoring')
 except ImportError:
     pass
+
+try:
+    from . import dataflow
+    __all__.append('dataflow')
+except ImportError:
+    pass
--- a/bigflow/dataflow.py
+++ b/bigflow/dataflow.py
@@ -1,0 +1,42 @@
+import typing
+
+from apache_beam import Pipeline
+from apache_beam.options.pipeline_options import PipelineOptions, GoogleCloudOptions
+
+from .workflow import Job, JobContext
+
+
+class BeamJob(Job):
+    def __init__(
+            self,
+            id: str,
+            driver_callable: typing.Callable[[Pipeline, JobContext, typing.Dict], None],
+            pipeline_options: PipelineOptions,
+            driver_arguments: typing.Optional[dict] = None,
+            wait_until_finish: bool = True):
+        self._id = id
+        self.driver_callable = driver_callable
+        self.driver_arguments = driver_arguments
+        self.pipeline_options = pipeline_options
+        self.wait_until_finish = wait_until_finish
+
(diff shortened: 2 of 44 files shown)


bigflow/dataflow.py
```diff
@@ -0,0 +1,42 @@
...
+            driver_arguments: typing.Optional[dict] = None,
+            wait_until_finish: bool = True):
+        self._id = id
+        self.driver_callable = driver_callable
+        self.driver_arguments = driver_arguments
+        self.pipeline_options = pipeline_options
+        self.wait_until_finish = wait_until_finish
+
+    def execute(self, context: JobContext):
+        pipeline_options = self._apply_logging(self.pipeline_options, context.workflow.workflow_id)
```
iwko, line 24: _apply_labels?

iwko:
Conflicts with master

bigflow/dataflow.py
anjensan, line 24: JobContext.worfkflow is `Optaionl[Workflow]`
Add 'if context.workflow' or we can change type of JobContext.workflow, which might be tricky however.

bigflow/__init__.py
```diff
@@ -26,3 +26,9 @@
     __all__.append('monitoring')
 except ImportError:
     pass
+
+try:
+    from . import dataflow
+    __all__.append('dataflow')
```
anjensan, line 32: IMO we should not add 'dataflow' to `__all__`
It is required only to support cases like:

```python
from bigflow import *
my_job = dataflow.BeamJob(...)
```

it is better to write
```python
import bigflow
my_job = bigflow.dataflow.BeamJob(...)
```
but the best option is to explicitly import needed module:

```python
from bigflow.dataflow import BeamJob
my_job = BeamJob(...)
```
or
```python
import bigflow.dataflow
my_job = bigflow.dataflow.BeamJob(...)
```

In such case there is no need to add anything to `bigflow/__init__.py` at all

bigflow/dataflow.py
```diff
@@ -0,0 +1,42 @@
...
+        pipeline = Pipeline(options=pipeline_options)
+        self.driver_callable(pipeline, context, **self.driver_arguments)
+        pipeline = pipeline.run()
+        if self.wait_until_finish:
+            pipeline.wait_until_finish()
+
+    @property
```
anjensan, line 31: There is no real need to add this property

bigflow/dataflow.py
```diff
@@ -0,0 +1,42 @@
...
+from apache_beam.options.pipeline_options import PipelineOptions, GoogleCloudOptions
+
+from .workflow import Job, JobContext
+
+
+class BeamJob(Job):
+    def __init__(
+            self,
+            id: str,
+            driver_callable: typing.Callable[[Pipeline, JobContext, typing.Dict], None],
```
anjensan, line 13: This function is used as `f(pipeline, context, **kwargs)`, not as `f(pipeline, context, kwargs)`
AFAIK in python 3.7 there is no way to express such callback type (with **kwargs)

anjensan (changes requested):

bigflow/dataflow.py
beczkowb: We apply labels to apply logging, so it's more meaningful to say "apply_logging".

bigflow/dataflow.py
beczkowb: switched from kwargs to a dict argument

bigflow/dataflow.py
beczkowb: true

bigflow/dataflow.py
beczkowb: done

bigflow/__init__.py
beczkowb: done

anjensan (approved):

PR #97, opened 2020-10-21
Update PySparkJob

anjensan:
- Migrate to new Job api
 - Extend JobContext (add workflow api / env)
 - Propagate logging config to driver / workers
    * Try to read log config from os.environ['bf_log_config'] during 'bigflow' initialization
 - Add basic scaffold
    * Introduce jinja2 templates to scaffolding


bigflow/dataproc.py
```diff
@@ -91,23 +94,26 @@ def _ensure_has_setup_file(self):
...
-        kwargs['runtime'] = runtime
-        
-        envs = {}
-        if self.env:
-            logger.debug("Will set env variable 'env' on pyspark driver to %r", self.env)
-            envs['env'] = self.env
+    def _prepare_driver_callable(self, context: bigflow.JobContext):
+        # Remove 'workflow' from context but keeps 'workflow_id'.
+        # In most cases Workflow is not pickleable as it keeps references to custom user jobs.
+        cleaned_context = context._replace(workflow=None)
```
beczkowb, line 107: Maybe you can add something like `PickableContext` which takes a standard context in the init. It clarifies the intention and allows you to avoid mutating the context parameter.

beczkowb (commented):
reviewed

bigflow/dataproc.py
anjensan: Idea was to pass this context (JobContext) to user-defined callback functiton.
Now it has signature: typing.Callback[JobContext]

I don't think there is real need to add something extra (some additional class) to such api, as 'workflow' is already Optional[]. As an option we can recreate 'workflow' object during driver unpickling, but IMO this is an overkill for now.

beczkowb: ok 👍

bigflow/scaffold/templates/new-project/{{project_name}}/pyspark/workflow.py.jinja
```diff
@@ -0,0 +1,40 @@
...
+    bucket_id=workflow_config.resolve_property('staging_bucket'),
+    gcp_project_id=workflow_config.resolve_property('gcp_project_id'),
+    gcp_region=workflow_config.resolve_property('gcp_region'),
+    
+    # Use the same set of dependencies as '{{project_name}}'
+    pip_packages=bigflow.resources.get_resource_absolute_path('requirements.txt', pathlib.Path(__file__)),
+)
+
+
+pyspark_demo_workflow = bigflow.Workflow(
```
beczkowb, line 35: Can you change the naming in this job to show what the job is doing? So here it's not `pyspark_workflow` but `pi_estimator`. Let's be consistent with the 2 other workflows.

beczkowb:
@anjensan looks good

bigflow/scaffold/templates/new-project/{{project_name}}/pyspark/workflow.py.jinja
anjensan: done

PR #102, opened 2020-10-23
Use default 'google.cloud.logging' Handler

anjensan:
- Send all log events from background thread
 - Pack logs in batches (up to 1k elements)
 - Handle log timestamps
 - Autodetect current GCP project id

### DIFF
--- a/bigflow/log.py
+++ b/bigflow/log.py
@@ -154,6 +154,8 @@ def init_logging(config: LogConfigDict, workflow_id: str, banner=True):
     # Disable logs from 'google.cloud.logging'
     gclogging_logger = logging.getLogger("google.cloud.logging")
     gclogging_logger.setLevel(logging.WARNING)
+    gclogging_logger.propagate = False
+    gclogging_logger.addHandler(logging.StreamHandler())
 
     # TODO: add formatter?
     root.addHandler(gcp_logger_handler)
--- a/bigflow/log.py
+++ b/bigflow/log.py
@@ -32,6 +32,7 @@ def create_gcp_log_handler(
 ):
     client = google.cloud.logging.Client(project=project_id)
     handler = google.cloud.logging.handlers.CloudLoggingHandler(client, name=log_name, labels=labels)
+    handler.addFilter(lambda r: not r.name.startswith("google.cloud.logging"))
     return handler
 
 
@@ -153,6 +154,10 @@ def init_logging(config: LogConfigDict, workflow_id: str, banner=True):
                ***********************************************************"""))
     gcp_logger_handler = create_gcp_log_handler(gcp_project_id, log_name, labels)
     gcp_logger_handler.setLevel(log_level or logging.INFO)
+
+    # Disable logs from 'google.cloud.logging'
+    gclogging_logger = logging.getLogger("google.cloud.logging")
+    gclogging_logger.setLevel(logging.WARNING)
 
     # TODO: add formatter?
     root.addHandler(gcp_logger_handler)
--- a/bigflow/log.py
+++ b/bigflow/log.py
@@ -2,7 +2,6 @@ import logging
 import typing
 import sys
 import uuid
-import traceback
 
 from textwrap import dedent
 
@@ -10,9 +9,6 @@ import bigflow
 
 import google.cloud.logging
 import google.cloud.logging.handlers
-
-from google.cloud import logging_v2
-from google.cloud.logging_v2.gapic.enums import LogSeverity
 
 from urllib.parse import quote_plus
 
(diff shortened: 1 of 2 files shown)


bigflow/log.py
```diff
@@ -8,8 +8,12 @@
 
 import bigflow
 
+import google.cloud.logging
+import google.cloud.logging.handlers
+
 from google.cloud import logging_v2
 from google.cloud.logging_v2.gapic.enums import LogSeverity
+
```
iwko, line 16: Some imports are not used

test/test_log.py
```diff
@@ -54,18 +56,13 @@ def test_should_create_correct_logging_link(self):
 
 
     def _assert_single_log_event(self, message_re, severity=None):
-        self.assertEqual(1, self.logging_client.write_log_entries.call_count)
-        calls = self.logging_client.write_log_entries.call_args_list[0][0]
-        le = calls[0][0]
-
-        self.assertEqual(le.resource.type, "global")
```
iwko, line 61: Why is it removed?

bigflow/log.py
```diff
@@ -158,15 +111,22 @@ def init_workflow_logging(workflow: 'bigflow.Workflow'):
...
     global _LOGGING_CONFIGURED
     if _LOGGING_CONFIGURED:
         import warnings
         warnings.warn(UserWarning("bigflow.log is already configured - skip"))
         return
     _LOGGING_CONFIGURED = True
 
-    gcp_project_id = config['gcp_project_id']
+    gcp_project_id = config.get('gcp_project_id') or _default_gcp_project()
```
iwko, line 129: In my opinion, we shouldnt do something like this.

bigflow/log.py
```diff
@@ -21,65 +25,14 @@
...
-        entry.json_payload.update(json)
-        entry.severity = LogSeverity[severity]
-        self.client.write_log_entries([entry])
+def create_gcp_log_handler(
+    project_id, 
+    log_name, 
+    labels,
+):
+    client = google.cloud.logging.Client(project=project_id)
+    handler = google.cloud.logging.handlers.CloudLoggingHandler(client, name=log_name, labels=labels)
```
iwko, line 34: Nice!

bigflow/log.py
```diff
@@ -21,65 +25,14 @@
 logger = logging.getLogger(__name__)
 
 
-class GCPLoggerHandler(logging.Handler):
```
iwko, line 24: Nice!

bigflow/log.py
anjensan: done

bigflow/log.py
anjensan: Could you explain how this might be possible?

test/test_log.py
anjensan: Resource/labels/etc is not attached to log-events.
It is handled underhood by the 'google.cloud.logging' library.

bigflow/log.py
beczkowb: @iwko ?

iwko: User may provide config without `gcp_project_id`. His default config in workflow config is `dev` and he expects that if he did not provide `--config` parameter logs will be sent to the `dev` project. But instead of that if `google.auth.default()` will return prod, project logs will go to prod project instead of dev.

anjensan: You are right, it may happen when users run jobs locally (not via cloud composer).

anjensan: Reverted.

bigflow/log.py
```diff
@@ -198,10 +148,13 @@ def init_logging(config: LogConfigDict, workflow_id: str, banner=True):
                Workflow logs (all runs): {workflow_logs_link}
                Only this run logs: {this_execution_logs_link}
                ***********************************************************"""))
-    
-    gcp_logger_handler = GCPLoggerHandler(gcp_project_id, log_name, labels)
+    gcp_logger_handler = create_gcp_log_handler(gcp_project_id, log_name, labels)
     gcp_logger_handler.setLevel(log_level or logging.INFO)
 
+    # Disable logs from 'google.cloud.logging'
```
iwko, line 154: As I understand we set "google.cloud.logging" level to WARNING to avoid sending it through gcp_logger_handler. But what if log_level from line 121 will be also WARNING? I think in line 156 we should setlevel as WARNING +1

anjensan: 'log_level' is used only to filter log messages. It is not related to level, at which messages may come from getLogger('google.cloud.logger')

In theory there are still might be cases, when some 'WARN' message might be generated and might cause recursive logging issue. So I've specified 'propagate = False' + attach separate console handler to this logger.

iwko (approved):

PR #104, opened 2020-10-26
Refactor, modules cleanup

anjensan:
This PR contain import-related changes:

  - Remove '__all__' from 'bigflow.bigquery', as names might go into clash with some variables from 'bigflow' namespace.
    It might encourage some users to write "from bigflow import *; from bigflow.bigquery import *"

  - Don't import 'bigquery' and 'monitoring' submodules in bigflow/__init__.py
     It makes library load slightly faster (important for CLI), easier to understand/read the sourcecode.
     Lazy-importing via "__getattr__" added in order to keep backward compatability.

  - Drop all "from __future__ import absolute_import" - noop for Py3

### DIFF
--- a/bigflow/__init__.py
+++ b/bigflow/__init__.py
@@ -1,6 +1,7 @@
 import os
 import sys
 import json
+import inspect
 import warnings
 import importlib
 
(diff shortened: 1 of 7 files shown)


bigflow/__init__.py
```diff
@@ -16,20 +17,21 @@
...
-    pass
+
+def __getattr__(name):
+    if name in {
+        'bigquery',
+        'monitoring',
+        'resources',
+    }:
+        importlib.import_module(f"bigflow.{name}")
+        msg = f"Don't use `bigflow.{name}` directly, add explicit import `import bigflow.{name}` instead"
```
beczkowb, line 30: I don't get it. Can you give an example of a wrong import, in that message?

bigflow/__init__.py
```diff
@@ -16,20 +18,27 @@
...
-except ImportError:
-    pass
-
-try:
-    from . import monitoring
-    __all__.append('monitoring')
-except ImportError:
-    pass
+
+# TODO(anjensan): Remove at 1.0
```
iwko, line 24: Why remove at 1.0?

anjensan: It is a hack to provide compatibility with previous versions of `bigflow`, so it might be better to remove it earlier.
We can keep this code for a while however.

bigflow/__init__.py
anjensan: I've changed message.
Also added small workaround to check way of importing. Now it should not print warning on statements like `from bigflow import dataproc`.

beczkowb: ok 👍

beczkowb (approved):
approved

iwko (approved):

PR #111, opened 2020-10-27
updating beam scaffold example

beczkowb:
Refactoring the beam example in scaffold so it uses BeamJob

### DIFF
--- a/bigflow/scaffold/scaffold_templates.py
+++ b/bigflow/scaffold/scaffold_templates.py
@@ -126,7 +126,7 @@ def count_words(p, target_method):
 '''
 beam_workflow_template = '''import logging
 
-import apache_beam as beam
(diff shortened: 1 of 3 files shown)


bigflow/scaffold/scaffold_templates.py
```diff
@@ -117,69 +126,36 @@ def count_words(p, target_method):
 '''
 beam_workflow_template = '''import logging
 
-from apache_beam.io import WriteToText
-
+import apache_beam as beam
```
iwko, line 129: Maybe import just Pipeline instead of whole beam?

iwko (approved):

bigflow/scaffold/scaffold_templates.py
beczkowb: done

PR #112, opened 2020-10-27
Beam utils doc update

beczkowb:


docs/technologies.md
```diff
@@ -83,6 +87,51 @@ google_cloud_options.staging_location = f"gs://{staging_location}"
...
+
+wordcount_workflow = bigflow.Workflow(
+    workflow_id="wordcount",
+    log_config={
+        'gcp_project_id': workflow_config['gcp_project_id'],
+        'log_level': 'INFO',
+    },
+    definition=[BeamJob(
+        id='wordcount_job',
+        driver_callable=wordcount_driver,
```
iwko, line 120: Why driver_callable? Why not job_callable?

beczkowb: To make it a bit similar to the dataproc job.

iwko: What we will gain thanks to this similarity?

beczkowb: consistent experience - if you know what is the driver in a dataproc job, then you might easier figure out what is a driver in a dataflow job

iwko: I am not sure if this is a good way. We introduce a new concept that is not commonly connected with a dataflow.

beczkowb: well, if you call this parameter `job_callable` it's confusing than helpful - job inside a job? should it implement the `execute` method? etc.
@anjensan what do you think?

beczkowb: #113

iwko (approved):

Issue #99, opened 2020-10-22
Removing the optional env variables prefix from Config class


beczkowb:
Currently, a user can set a custom prefix to the environmental variables in the Config class. The problem is that in many places we assume that the prefix has the default value. Currently, it makes no sense to keep that as customizable, it needs to be removed.

anjensan:
Feature was removed from documentation of `bigflow.Config`
It is still used internally however => we are going to keep it for now. Close this issue.

PR #115, opened 2020-10-28
Update 'bigflow.Config'

anjensan:


bigflow/configuration.py
```diff
@@ -1,44 +1,47 @@
...
         self.default_env_name = None
-        self._update_default_env_name(name, is_default)
+        self.configs = {}
 
-        self.environment_variables_prefix = environment_variables_prefix
+        self._wrap_cached_methods()
+        self.add_configuration(name, properties, is_default)
 
-        self._check_docker_repository(properties)
+    def _wrap_cached_methods(self):
```
beczkowb, line 37: Why do we need to cache the resolve method results?

anjensan: Just a small optimization.

Might be useful if someone heavily uses 'resolve_property' (which triggers string interpolation of the whole config; which is O(N**2)). So this code has complexity O(N**3)
  print([c.resolve_property(x) for x in c.resolve()])

Directly caching 'resolve_property' is less beneficial because it will recall 'resolve()' for each property. Complexity of the above code will be O(N**2), not O(N)

beczkowb: Well, typically N should be like 5, 10, 50 maybe 100, so 100*100*100 is not a problem. But, on the other hand, who knows how a user can use this class. I leave it to @iwko and @bartoszwalacik .

beczkowb (commented):
@anjensan

bigflow/configuration.py
iwko: I think for this moment it is overengineering to use caching

anjensan: Some benchmarking - [10...1000] properties, 1 subst var, each property resolved once via 'resolve_property'
https://gist.github.com/anjensan/6ee025910af91e56e3825b8dd917ae5a

for 100 items the difference is quite small - just 0.28 vs 0.003 seconds
for 500 - slightly bigger - 35 vs 0.7

PS: actually I made an error. Complexity of single call is `O(N**3 * `complexity of str.replace)`

bartoszwalacik: I agree with @iwko , we don't need cache for config props

anjensan: Reverted.

Maybe add some additional comment to documentation that it is not recommended to pass 'config' into a job and invoke 'resolve_property' from it (may be in some loop), and better is to resolve all properties outside the job?

iwko (approved):

bigflow/configuration.py
beczkowb: We should solve the problem instead of informing a user how to deal with it. And because we don't want to solve the problem right now, let's also skip the recommendation.

beczkowb (approved):
approved

bigflow/configuration.py
anjensan: IMO problem should be resolved OR the fact of this problem should be tracked (create a bug?) OR we should inform user that we have that problem OR problem should be clarified as "wont-fix"

Spoke with @bartoszwalacik - agreed to slightly change `resolve()` in such way, that user will not be enforced to use 'resolve_property' & keep Config without caching. In this scenario performance of 'resolve_property' doesn't matter (as almost nobody will use it).

anjensan:
Fixes #99

Issue #103, opened 2020-10-25
"Dotify" artefacts


beczkowb:
Currently, not all of the deployment artifacts have a `.` prefix. We want to apply the `.` prefix to all of them.

PR #116, opened 2020-10-28
adding "." as a image dir prefix

beczkowb:
#103

### DIFF
--- a/bigflow/scaffold/scaffold_templates.py
+++ b/bigflow/scaffold/scaffold_templates.py
@@ -385,8 +385,8 @@ dmypy.json
 # Pyre type checker
 .pyre/
 .dags
-image
+.image
 .dags/
-image/
+.image/
 .idea'''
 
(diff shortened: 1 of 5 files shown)


iwko (approved):

bigflow/scaffold/scaffold_templates.py
```diff
@@ -385,8 +385,8 @@ class FakeFile(object):
 # Pyre type checker
 .pyre/
 .dags
-image
+.image
 .dags/
-image/
+.image/
```
anjensan, line 390: Maybe '/image/' (don't ignore 'image' directory/file outside the project root)
Also I don't think there is need to add both 'image' and 'image/'

anjensan (approved):
Approve with one comment.

bigflow/scaffold/scaffold_templates.py
beczkowb: true, done

PR #118, opened 2020-10-29
DeploymentConfig with the optional env_variables_prefix parameter, re…

beczkowb:
…moving that parameter from the Config class

### DIFF
--- a/bigflow/configuration.py
+++ b/bigflow/configuration.py
@@ -13,7 +13,7 @@ class EnvConfig(typing.NamedTuple):
 
 def current_env():
     """Returns current env name (specified via 'bigflow --config' option)"""
-    return os.environ.get('bf_env')
+    return os.environ.get(f'{DEFAULT_CONFIG_ENV_VAR_PREFIX}env')
 
 
 class Config:
@@ -27,6 +27,7 @@ class Config:
         self.default_env_name = None
         self.configs = {}
         self.add_configuration(name, properties, is_default)
+        self.environment_variables_prefix = DEFAULT_CONFIG_ENV_VAR_PREFIX
 
     def __str__(self):
         return "\n".join(map(self.pretty_print, self.config.keys()))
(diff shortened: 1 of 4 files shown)


iwko:
Why you've created separate class for DeploymentConfig?

beczkowb:
@iwko This is a solution for #99 . Environment variable prefix works only for the `bf deploy` command (because we need the prefix for bamboo). In any other case we assume that the prefix == `bf_`. So it's misleading for a user to have env prefix parameter in the Config class.

iwko (approved):

bigflow/configuration.py
```diff
@@ -111,3 +114,16 @@ def _resolve_placeholders(self, value, variables: dict):
...
             return value
 
+
+class DeploymentConfig(Config):
+    def __init__(self,
+                 name: str,
+                 properties: dict,
+                 is_master: bool = True,
+                 is_default: bool = True,
+                 environment_variables_prefix: str = DEFAULT_CONFIG_ENV_VAR_PREFIX):
```
anjensan, line 124: nit: consts are not actual consts in PY, so someone might change variable 'DEFAULT_CONFIG_ENV_VAR_PREFIX' and will not be reflected on this class/parameter. Maybe `None` OR drop the 'const variable'.

anjensan (approved):
In general approve.

*Maybe* just add this as a "hidden feature" to base 'Config' class
In fact this feature is already here, as you can write
```python
c = bigflow.Config("name", {})
c.environment_variables_prefix = "foo_"
```

It is not documented neither for Config not for DeployConfig...
So it is hidden from user in any case.

bigflow/configuration.py
beczkowb: done

beczkowb:
@anjensan I will update the docs later, now we want to finish the code so we can publish RC1.

PR #120, opened 2020-10-29
Return all params from Config.resolve()

anjensan:


bigflow/configuration.py
```diff
@@ -27,43 +22,64 @@ def __init__(self,
...
+        res = {
+            key: self._resolve_placeholders(value, properties_with_placeholders)
+            for key, value in properties_with_placeholders.items()
+        }
 
-        string_properties = {k: v for (k, v) in properties_with_placeholders.items() if isinstance(v, str)}
-        string_properties['env'] = env_config.name
+        if 'env' not in env_config:
+            # For backward compatability - don't show "magic" 'env' variable to user
+            res.pop('env', None)
```
beczkowb, line 73: It's a useful little thing (we use it in our projects).

anjensan: It is still here (so might be used during interpolation), but excluded from '.resolve()'
I doubted about this, so probably we should not overcomplicate this and always add 'env' property with current config name (even when it is not specified via os.environ).
wdyt?

beczkowb: Sounds like a consistent solution, agree

PR #121, opened 2020-10-29
Fix dataproc - patch os.environ before loading of user code

anjensan:
During unpickling of dataproc driver callable user code will be
imported. This code might analyse `os.environment` via Config().resolve().


beczkowb (approved):
approved

PR #129, opened 2020-11-02
docs fixes part1

beczkowb:
part 1

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -135,7 +135,7 @@ The `build-dags` command takes two optional parameters:
 
 * `--start-time` &mdash; the first [runtime](workflow-and-job.md#the-runtime-parameter)
   of your workflows. If empty, a current hour (`datetime.datetime.now().replace(minute=0, second=0, microsecond=0)`)
-  is used for hourly workflows and `datetime.date.today()` for daily workflows.
+  is used.
 * `--workflow` &mdash; leave empty to build DAGs from all workflows.
    Set a workflow Id to build a selected workflow only.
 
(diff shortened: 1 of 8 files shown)


docs/cli.md
```diff
@@ -135,7 +135,7 @@ The `build-dags` command takes two optional parameters:
 
 * `--start-time` &mdash; the first [runtime](workflow-and-job.md#the-runtime-parameter)
   of your workflows. If empty, a current hour (`datetime.datetime.now().replace(minute=0, second=0, microsecond=0)`)
-  is used for hourly workflows and `datetime.date.today()` for daily workflows.
+  is used.
```
bartoszwalacik, line 138: why removing this?

docs/configuration.md
```diff
@@ -303,4 +308,22 @@ prod config:
 {'my_project': 'I_am_prod'}
 dev config:
 {'my_project': 'I_am_dev'}
+```
+
+## Deployment config
+
+`DeploymentConfig` is a subclass of the `Config` class with one additional, optional argument – `environment_variables_prefix`.
+Contrary to the `Config` class, the `DeploymentConfig` allows you to set a custom environment variables name prefix (overriding the default `bf_` value).
+It's useful in CI/CD systems like Bamboo.
```
bartoszwalacik, line 317: CI/CD servers

bartoszwalacik: btw why changing this prefix would be useful on CI/CD server?

docs/cli.md
beczkowb: Because currently, the default start-time value is always date time. There is a user-defined "start_time_factory", which produces a proper start-time.

docs/configuration.md
beczkowb: done

docs/cli.md
bartoszwalacik: so maybe this should also be updated ?
https://github.com/allegro/bigflow/blob/master/bigflow/cli.py#L325

beczkowb: 👍 done

bartoszwalacik (approved):

PR #147, opened 2020-11-18
Relax deps version for 'bigquery' extras - allow minor upgrades

anjensan:


beczkowb (approved):

iwko:
Why we want this to change?

iwko (approved):

anjensan:
@iwko Some our clients (actually our own project) requires google-cloud-core==1.4.3
At the same time it is not a good practice to force rest of our clients to upgrade (even if this upgrade is minor).

Better is to provide a way to pin all libs versions at client side (actually I'm working on a PR for this).

Issue #98, opened 2020-10-22
Use 'pip-tools' in scaffolded projects


anjensan:
It is a good practice to recommend to our clients 'pip-tools' (or other similar tool) for dependencies management.
We should add this to our 'scaffold'.

PR #151, opened 2020-11-20
Clean dependencies

anjensan:
- Don't include 'requirements.txt' into 'data' files (there is no need to do this)
- Allow user to do minor-upgrade for dependencies
- This PR don't require to use 'pip-tools', but synergises with #98


requirements/base.txt
```diff
@@ -1,10 +1,9 @@
...
\ No newline at end of file
+google-cloud-storage>=1.32,<2
+google-auth>=1.20,<2
+
+better-setuptools-git-version>=1.0.5,<2
+unittest-xml-reporting>=3.0.2,<4
+jinja2>=2.10,<3
+
+wheel>=0.35
+setuptools>=49
```
beczkowb, line 9: >=49,<50?

setup.py
```diff
@@ -30,16 +30,6 @@ def read_requirements(name):
     url="https://github.com/allegro/bigflow",
     packages=setuptools.find_packages(exclude=('test', 'e2e')),
     include_package_data=True,
-    data_files=[
```
iwko, line 33: Why is it deleted?

requirements/base.txt
```diff
@@ -1,10 +1,9 @@
...
-jinja2<2.11.0,>=2.10.1
\ No newline at end of file
+google-cloud-storage>=1.32,<2
+google-auth>=1.20,<2
+
+better-setuptools-git-version>=1.0.5,<2
+unittest-xml-reporting>=3.0.2,<4
+jinja2>=2.10,<3
+
+wheel>=0.35
```
beczkowb, line 8: same here, shouldn't we add the upper bound?

test/requirements.txt
```diff
@@ -1,5 +1,4 @@
-apache-airflow==1.10.11
-apache-beam==2.23.0
-six==1.14.0
-dill==0.3.0
+apache-airflow>=1.10
```
iwko, line 1: In this case, we allow installing 2.0 airflow which will be probably incompatible

test/requirements.txt
```diff
@@ -1,5 +1,4 @@
...
 nose==1.3.7
+mock>=2.0
```
iwko, line 3: Same case as with airflow

requirements/bigquery_extras.txt
```diff
@@ -1,5 +1,4 @@
-google-cloud-core~=1.4
+# TODO: Upgrade google-cloud-bigquery to ~2.0
 google-cloud-bigquery~=1.28
-google-auth>=1.20.1,<2.0dev
-google-apitools>=0.5.31,<0.6
-pandas~=1.1
\ No newline at end of file
+pandas>=1.1,<2
+six>=1.14
```
iwko, line 4: Same case as with airflow

setup.py
anjensan: Because it is not needed nor used.
Now all *.txt files from bigflow are installed into $VIRTUAL_ENV/requirements
AFAIK nobody uses them.

requirements/base.txt
anjensan: AFAIK pinning 'setuptools' might break some 'venv' related tools (but can't find clear details about this).
Also 'setuptools' are used to install package. If version of installed setuptools differ from version in 'bigflow' - package might be reinstalled during it's usage.

anjensan: BTW: 'pip-tools' doesn't include version of 'setuptools' into generated 'requirements.txt' file on purpose.
https://github.com/jazzband/pip-tools/blob/61eb515b430a57e71f0bf768049e6a1f23d3c7c8/piptools/utils.py#L17

beczkowb: ok, closing this comment

test/requirements.txt
anjensan: It was done on purpose - validate that bigflow-generated DAGs are compatible with latest stable version of 'airflow' (this dependency used only for development/ci). As we actually don't have any control on what version of Airflow is used by our clients.

requirements/bigquery_extras.txt
anjensan: It is *extremely* unlikely, that Six will recieve any major update ever given that Python 2.x is officially unsupported.

test/requirements.txt
anjensan: I've changed to pinned version for all test dependencies except airflow.

requirements/base.txt
anjensan: AFAIK 'wheel' is also preinstalled even on fresh 'venv'. Also it is not directly used by 'bigflow', so it is better to allow 'setuptools' apply it's own restrictions.

beczkowb (approved):
From me - approved, @iwko ?

iwko:
LGTM

iwko (approved):

PR #157, opened 2020-11-24
Refactoring & cleanup

anjensan:


bigflow/commons.py
```diff
@@ -24,23 +21,6 @@ def run_process(cmd, **kwargs):
     return result_output
 
 
-def resolve(path: Path):
```
beczkowb, line 27: It's a part of the API visible in the docs, scaffold, our source code. If you want to remove it (in another PR), do it by:

* updating examples from docs
* updating scaffold
* updating out in-house starter
* adding "deprecated" message
* ...?

bigflow/build.py
```diff
@@ -33,16 +34,16 @@ def run_tests(build_dir: Path, test_package: Path):
...
 
 def build_docker_image(project_dir: Path, tag: str):
     print('Building a Docker image. It might take a while.')
-    run_process(f'docker build {resolve(project_dir)} --tag {tag}')
+    run_process(f'docker build {project_dir} --tag {tag}')
 
 
 def build_dags(
-        root_package,
+        root_package: Path,
```
beczkowb, line 46: it would be nice if you would add typing in the whole module :)

bigflow/scaffold/templating.py
```diff
@@ -10,7 +10,7 @@
 logger = logging.getLogger(__name__)
 
 
-class SkipThisTemplate(BaseException):
+class SkipThisTemplate(Exception):
```
beczkowb, line 13: Can't we use a standard library exception instead of `SkipThisTemplate`?

beczkowb (changes requested):
The `resolve` function is a big deal, the rest of the comments are minors.

bigflow/build.py
```diff
@@ -102,7 +103,7 @@ class BuildCommand(distutils.cmd.Command):
         ]
 
         def initialize_options(self) -> None:
-            self.start_time = now()
+            self.start_time = datetime.now().strftime("%Y-%m-%d %H:00:00")
```
iwko, line 106: I prefer the previous solution

bigflow/build.py
anjensan: done

bigflow/scaffold/templating.py
anjensan: This is a custom exception to prevent file from being rendered.
It is not an indicator of an error, but is more like "hacky" control-flow (like StopIteration).
We can't use any default error here as it catched by 'templating.py' module.

bigflow/commons.py
anjensan: Added back 'resolve' with deprecation warning. Updated docs/scaffold.

bigflow/build.py
anjensan: IMO explicit is much better. It is extremely unclear what is going here:
 - what type does 'now' return?
 - does it include date, time, or "hours without minutes and seconds"
As an option we can refactor datetime format into some named const, but I still prefer explicit format in this case.

beczkowb (approved):

PR #161, opened 2020-11-26
dataflow timeout

iwko:


bigflow/dataflow.py
```diff
@@ -12,31 +13,35 @@
 
 class BeamJob(Job):
     def __init__(
```
beczkowb, line 15: Don't change the order of the arguments because you break the API backward compatibility.

bigflow/dataflow.py
```diff
@@ -12,31 +13,35 @@
...
-        id: str,
-        entry_point: typing.Callable[[Pipeline, JobContext, dict], None],
-        pipeline_options: PipelineOptions,
-        entry_point_arguments: typing.Optional[dict] = None,
-        wait_until_finish: bool = True,
+            self,
+            id: str,
+            entry_point: typing.Callable[[Pipeline, JobContext, dict], None],
+            pipeline_options: PipelineOptions,
+            pipeline: Type[Pipeline],
```
beczkowb, line 20: Assuming we want this parameter for tests:

1. It should be optional
2. It should be a pipeline object, not a pipeline class (so, for example, a user can use `with TestPipeline() as p` clause)
3. It should be called `test_pipeline`, to strongly suggest the purpose

iwko: ad1. Changed, see now.
ad2. If we pass an instance of the Pipeline class then we will not be able to apply_logging inside the execute method. In that case, the user has to provide logging by himself.
ad3 I do not like `test_pipeline` name,  maybe I will add info about test purpose in docs?

beczkowb: ad2. and thats ok, because it's for testing purposes
ad3. it's for testing, so why not `test_pipeline`?

anjensan: ad2. That is ok, but I'd add some checks like "assert not pipeline or not pipeline_options"
So user can specify either "pipeline_options" OR pipeline object.

ad3. I also don't like `test_pipeline` name, but It might protect our users from creating pipelines manually in the prod code, which is good. So I'd vote for "ugliest name" (eg `test_pipeline`).

test/test_dataflow.py
```diff
@@ -1,51 +1,78 @@
...
         words_input = pipeline | 'LoadingWordsInput' >> beam.Create(driver_arguments['words_to_count'])
-        self.result = (words_input
-                       | 'FilterWords' >> beam.Filter(lambda w: w in driver_arguments['words_to_filter'])
-                       | 'MapToCount' >> beam.Map(lambda w: (w, 1))
-                       | 'GroupWords' >> beam.GroupByKey()
-                       | 'CountWords' >> beam.ParDo(CountWordsFn()))
+        words_input | 'FilterWords' >> beam.Filter(lambda w: w in driver_arguments['words_to_filter']) \
+        | 'MapToCount' >> beam.Map(lambda w: (w, 1))\
+        | 'GroupWords' >> beam.GroupByKey()\
+        | 'CountWords' >> self.saver
```
anjensan, line 53: nit: AFAIK it is not recommended to use '\' line continuation. https://www.python.org/dev/peps/pep-0008/#maximum-line-length

test/test_dataflow.py
```diff
@@ -1,51 +1,78 @@
...
+        super().__init__()
+        self.save = save
+
     def process(self, element, *args, **kwargs):
         word, count = element
+        self.save.counter[word] += len(count)
         yield word, len(count)
 
 
+class Save(object):
```
anjensan, line 26: Why not to use just `defaultdict(int)` instead of this `wrapping` class?

iwko:
HTT:
- create job with low value of timeout (for example 240000 ms )
- run job locally
- job should timeout after 4 mins

test/test_dataflow.py
iwko: In that case `self.assertEqual(driver.saver.save.counter, {'valid': 2, 'word': 1})` is never True. driver.saver.save.counter has no values,

test/test_dataflow.py
iwko: This is how we do it in our scripts

test/test_dataflow.py
```diff
@@ -66,4 +88,147 @@ def test_should_run_beam_job(self):
...
+        # when
+        count_words.run('2020-01-01')
+
+        # then
+        self.assertEqual(cancel_mock.call_count, 0)
+        wait_until_finish_mock.assert_called_with(3600000)
+
+    @patch.object(RunnerResult, 'is_in_terminal_state', create=True)
+    @patch.object(RunnerResult, 'cancel')
+    def test_should_run_beam_job_without_timeout_if_wait_until_finish_disabled(self, cancel_mock, is_in_terminal_state_mock):
```
beczkowb, line 152: You can just validate is that situation occurred and throw the ValueError (if you disabled wait_until_finish, then you can't use the timeout).

beczkowb (changes requested):
needs work

test/test_dataflow.py
anjensan: 'beam' official documentation [uses](https://github.com/apache/beam/blob/master/sdks/python/apache_beam/examples/wordcount_minimal.py#L112) different style (with extra `()` but without `\`).
Maybe we should consider using the "official" approach ?

test/test_dataflow.py
anjensan: Well, looks like it is effectively a global variable (aka 'static' field).
So you could replace `Save.counter` with something like `GLOBAL_COUNTER`, which is not the best approach, but a little more readable.

test/test_dataflow.py
iwko: Changed

test/test_dataflow.py
iwko: Changed

beczkowb (approved):
approved, @anjensan ?

test/test_dataflow.py
iwko: I will stay with Save class

anjensan: Well, maybe at least provide more verbose name/comment, because it is quite hard to understand what is `Save`

anjensan (changes requested):

bigflow/dagbuilder.py
```diff
@@ -43,7 +45,7 @@ def generate_dag_file(workdir: str,
             'start_date': {start_date_as_str},
             'email_on_failure': False,
             'email_on_retry': False,
-            'execution_timeout': datetime.timedelta(minutes=180),
+            'execution_timeout': datetime.timedelta(milliseconds={execution_timeout}),
```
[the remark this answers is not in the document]
iwko: Good point

bigflow/dataflow.py
```diff
@@ -5,38 +5,56 @@
...
+            id: str,
+            entry_point: typing.Callable[[Pipeline, JobContext, dict], None],
+            pipeline_options: PipelineOptions = None,
+            entry_point_arguments: typing.Optional[dict] = None,
+            wait_until_finish: bool = True,
+            execution_timeout: int = DEFAULT_EXECUTION_TIMEOUT,
+            test_pipeline: Pipeline = None,
+            pipeline_level_execution_timeout_shift: int = DEFAULT_PIPELINE_LEVEL_EXECUTION_TIMEOUT_SHIFT
     ):
+        if (test_pipeline and pipeline_options) or (not test_pipeline and not pipeline_options):
```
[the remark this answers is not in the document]
iwko: Nice!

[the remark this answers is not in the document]
iwko: Nice! :)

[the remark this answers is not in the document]
iwko: Done

bigflow/dagbuilder.py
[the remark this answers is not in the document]
iwko: Done

bigflow/commons.py
```diff
@@ -4,11 +4,11 @@
 import logging
 
 from pathlib import Path
-from datetime import datetime
-
 
 logger = logging.getLogger(__name__)
 
+DEFAULT_EXECUTION_TIMEOUT = 10800000  # 3 hours
```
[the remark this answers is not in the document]
iwko: Moved

anjensan (approved):

Issue #130, opened 2020-11-03
Allow user to change timeout for a single Job


anjensan:
Currently 'execution_timeout' is hardcoded into generated DAG.
Add ability to specify different timeout for Jobs. Possible solution: add new field to base Job.

anjensan:
Need to add validation.
User can specify any invalid value (string, timedelta object etc) and invalid DAG will be generated.
Expected: DAG building should fail.

PR #163, opened 2020-12-02
#130 job timeout

iwko:


iwko:
HTT:
- create a job with custom value of job_execution_timeout for example 240000 ms (4 minutes)
- build and deploy job
- job should timeout on airflow after 4 mins

extra E2E test for ambitious:
- Set airflow timeout for example for 6 mins, dataflow for 4.  After dataflow is timed out airflow should return failed task

bigflow/commons.py
```diff
@@ -4,11 +4,11 @@
 import logging
 
 from pathlib import Path
-from datetime import datetime
-
 
 logger = logging.getLogger(__name__)
 
+DEFAULT_JOB_EXECUTION_TIMEOUT = 3600000  # 1 hour
```
beczkowb, line 10: It used to be longer, keep the same execution timeout as it was to avoid breaking users' workflows.

bigflow/dataflow.py
```diff
@@ -18,8 +18,9 @@ def __init__(
             pipeline_options: PipelineOptions = None,
             entry_point_arguments: typing.Optional[dict] = None,
             wait_until_finish: bool = True,
-            execution_timeout: int = 3600000,
-            test_pipeline: Pipeline = None
+            execution_timeout: int = DEFAULT_JOB_EXECUTION_TIMEOUT - 120000,  # minus two minutes to allow airflow report to user timeout of BeamJob
```
beczkowb, line 21: Why there are 2 timeout parameters? (execution_timeout and job_execution_timeout)

beczkowb:
You should add the timeout parameter to the main Job class - `bigflow.Job`

test/test_dataproc.py
```diff
@@ -347,3 +343,16 @@ def test_should_initialize_logging_on_workers(self):
         # then
         self.assertIn('spark.executorEnv.bf_log_config', props)
         self.assertIn('spark.executorEnv.bf_workflow_id', props)
+
+    def test_should_initialize_job_with_default_job_execution_timeout(self):
```
beczkowb, line 347: what about testing non default execution timeout in each Job class?

bigflow/commons.py
iwko: Ok

bigflow/dataflow.py
iwko: job_execution_timeout is used in DAG and passed to KubernetesPodOperator. It is used to control timeout at airflow side. execution_timeout is timeout at dataflow

test/test_dataproc.py
iwko: Missing, will ad

bigflow/commons.py
iwko: Set to None as default

test/test_dataproc.py
iwko: Added

bigflow/commons.py
beczkowb: Why None? I think it should be the same value as it used to be in the previous version.

beczkowb: And if you set None as default then what's the final default timeout (it can't be just None)?

bigflow/bigquery/job.py
```diff
@@ -14,12 +16,14 @@ def __init__(self,
                  id=None,
                  retry_count=DEFAULT_RETRY_COUNT,
                  retry_pause_sec=DEFAULT_RETRY_PAUSE_SEC,
+                 job_execution_timeout=DEFAULT_JOB_EXECUTION_TIMEOUT,
```
beczkowb, line 19: I think `execution_timeout` is a better name, the `job_` prefix is redundant (because you set the execution_timeout on a job) and not consistent with the rest of the parameters (there is, for example, `retry_count` param, not `job_retry_count`.

bigflow/dataflow.py
```diff
@@ -18,8 +18,9 @@ def __init__(
             pipeline_options: PipelineOptions = None,
             entry_point_arguments: typing.Optional[dict] = None,
             wait_until_finish: bool = True,
-            execution_timeout: int = 3600000,
-            test_pipeline: Pipeline = None
+            execution_timeout: int = None,
+            test_pipeline: Pipeline = None,
+            job_execution_timeout: int = DEFAULT_JOB_EXECUTION_TIMEOUT
```
beczkowb, line 23: Without any context, a user has no clue why there are 2 execution timeouts. I suggest using a single parameter - `execution_timeout` and an additional parameter, like `pipeline_level_execution_timeout_shift`, so the final execution timeout will be `execution_timeout - pipeline_level_execution_timeout_shift`.

@anjensan what do you think?

beczkowb (changes requested):
needs work

bigflow/bigquery/job.py
```diff
@@ -1,7 +1,9 @@
 import bigflow
 
 from inspect import getargspec
+
 from .dataset_manager import create_dataset_manager
+from ..commons import DEFAULT_JOB_EXECUTION_TIMEOUT
```
anjensan, line 6: Use `import bigflow.commons` instead

bigflow/commons.py
```diff
@@ -4,11 +4,11 @@
 import logging
 
 from pathlib import Path
-from datetime import datetime
-
 
 logger = logging.getLogger(__name__)
 
+DEFAULT_JOB_EXECUTION_TIMEOUT = None
```
anjensan, line 10: Do we really need this const?
IMO it is not the best place for it anyway.

bigflow/dataflow.py
```diff
@@ -5,7 +5,7 @@
 from apache_beam.options.pipeline_options import PipelineOptions, GoogleCloudOptions
 
 from bigflow.workflow import Job, JobContext
-
+from .commons import DEFAULT_JOB_EXECUTION_TIMEOUT
```
anjensan, line 8: Please, no relative imports.

bigflow/dagbuilder.py
```diff
@@ -75,15 +75,16 @@ def build_dag_operator(workflow_job, dependencies):
...
 """.format(job_var=job_var,
           task_id=task_id,
           docker_image = docker_repository+":"+build_ver,
           bf_job= workflow.workflow_id+"."+job.id,
           root_folder=root_package_name,
           retries=job.retry_count if hasattr(job, 'retry_count') else 3,
-          retry_delay=job.retry_pause_sec if hasattr(job, 'retry_pause_sec') else 60))
-
+          retry_delay=job.retry_pause_sec if hasattr(job, 'retry_pause_sec') else 60,
+          execution_timeout=timedelta(milliseconds=job.job_execution_timeout) if job.job_execution_timeout else None))
```
anjensan, line 87: We can use `datetime.timedelta` as type for `execution_timeout.
If it can't be `timedelta` for some reason (eg it comes from cmd-line / json etc), than it is better to use 'float seconds' (which is consistent with `time.time`, `time.sleep` and other functions).

docs/technologies.md
```diff
@@ -133,9 +133,10 @@ a pipeline for a driver. One of the`pipeline_options`, `test_pipeline` must be p
 * The `entry_point_arguments` parameter should be a dictionary. It can be used in a entry point as a configuration holder.
 * The `wait_until_finish` parameter of bool type, by default set to True. It allows to timeout Beam job.
 * The `execution_timeout` parameter of int type. If `wait_until_finish` parameter is set to True it provides an interval after
-which Beam job will be considered as timed out. The default value is 3600000 ms (1 hour).
+which Beam job will be considered as timed out. The default value is 1 hour minus two minutes.
```
anjensan, line 136: Why default timeout is 58 minutes?

test/test_dagbuilder.py
```diff
@@ -112,7 +112,8 @@ def test_should_generate_DAG_file_from_workflow_with_hourly_scheduling(self, get
     is_delete_operator_pod=True,
     retries=10,
     retry_delay=datetime.timedelta(seconds=20),
-    dag=dag)
+    dag=dag,
+    execution_timeout=None)
```
anjensan, line 116: nit: it is possible to add trailing `,` after last argument, which results into smaller diffs when you add/remove/reorder named arguments.

test/test_dataflow.py
```diff
@@ -90,6 +91,10 @@ def test_should_run_beam_job(self, is_in_terminal_state_mock):
             driver.pipeline._options.get_all_options()['labels'],
             ['workflow_id=count_words'])
 
+        # and sets default value for job_execution_timeout and execution_timeout
+        self.assertEqual(job.job_execution_timeout, DEFAULT_JOB_EXECUTION_TIMEOUT)
```
anjensan, line 95: nit: `assertIsNone`

bigflow/workflow.py
```diff
@@ -104,6 +104,7 @@ class Job(abc.ABC):
     id: str
     retries: int = 3
     retry_delay: float = 60
+    job_execution_timeout: int
```
anjensan, line 107: Add default value for `job_execution_timeout` here.
Otherwise attribute is missing, (type hint doesn't declare actual attribute).

anjensan (changes requested):

bigflow/dataflow.py
anjensan: +1 for having `execution_timeout` as main entry for users.
IMO `pipeline_level_execution_timeout_shift` is not really needed to be exposed to user (but we might still have such option as attribute of 'Job' objects).

bigflow/commons.py
iwko: > Why None? I think it should be the same value as it used to be in the previous version.
And this is value from previous version. Default values in dataflow and kubernetesPodOperator are None.

bigflow/bigquery/job.py
iwko: I've added job_ prefix because this timeout is passed to airflow job

bigflow/dataflow.py
iwko: Ok, will change

bigflow/bigquery/job.py
iwko: removed

bigflow/commons.py
iwko: Removed

bigflow/dataflow.py
iwko: Done

docs/technologies.md
iwko: Fixed

test/test_dataflow.py
iwko: Done

bigflow/workflow.py
iwko: Done

bigflow/dagbuilder.py
iwko: Straight int is simplier for user imo

bigflow/commons.py
beczkowb: True, previously the default timeout was inherited from the DAG settings.

iwko:
>You should add the timeout parameter to the main Job class - bigflow.Job @iwko

@beczkowb see now

bigflow/workflow.py
```diff
@@ -104,6 +104,7 @@ class Job(abc.ABC):
     id: str
     retries: int = 3
     retry_delay: float = 60
+    execution_timeout: int = None
```
beczkowb, line 107: Provide the backward compatible default value explicitly so a user knows what is the default value.

beczkowb (approved):

PR #164, opened 2020-12-03
Change 'setup.py' for bigflow projects

anjensan:
Internal changes:
 * Automatically generates 'sdist' package & include it as resource into .whl & .egg packages
 * Add a few internal APIs:
   - Unpack embedded 'sdist' & compile egg/whl/sdist from local source OR pip-installed package
   - Read any parameters from 'setup.py' (including package name)
   - Infer project name and 'setup.py' location
 * Automatically migrate project structure when used for old project
 * Generated egg/whl/sdist packages requires the same set of dependencies as original bf project
   - It may add some extra deps, but is much less error-prone approach
   - There is no need to explicitly read/parse/process 'requirements.txt' for dataflow/dataproc
 * Check subprocesses exit code
 * Update dataflow/dataproc code/templates

Public API changes:
  * rename `project_setup.py` to `setup.py`
  * single `bigflow.build.setup(..)` as entry point of bf building
  * deprecate functions from 'bigflow.build'
  * add `pyproject.toml` & `MANIFEST.in`
  * magic variable `PROJECT_NAME` is no longer required
  * some functions at 'bigflow.resources' are deprecated (no longer needed)


anjensan:
BTW: I don't like new module names `bf.build.reflect` / `bf.build.dev` etc. Any suggestions are welcome :)

anjensan:
More tests will be added soon, but code is ready for pre-review.

bigflow/build/dev.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+def find_setuppy(directory: typing.Union[None, Path, str] = None) -> Path:
+    """Find location of project setup.py.
+
+    Scan provided directory and its parents, searching for any `setup.py`
+    Scanning doesn't escape user home directory."""
+
+    directory = Path(directory or Path.cwd())
+    while directory != directory.parent and directory != Path.home():
```
beczkowb, line 73: Looks similar to `bigflow.resources.find_file`, use it if you can.

bigflow/build/dist.py
```diff
@@ -1,43 +1,36 @@
...
-    'auto_configuration',
-    'default_project_setup'
-]
+import bigflow.cli
+import bigflow.resources
+import bigflow.dagbuilder
+import bigflow.version
+import bigflow.build.pip
+import bigflow.build.dev
+import bigflow.commons as bfc
```
beczkowb, line 27: bfc -> `bf_commons` ?

bigflow/build/dist.py
```diff
@@ -267,6 +307,11 @@ def project_setup(
     setup(project_setup(**auto_configuration('my_super_project')))
     '''
 
+    _maybe_dump_setup_params({
+        'name': project_name,
+    })
+    bigflow.build.pip.maybe_recompile_requirements_file(project_requirements_file)
```
beczkowb, line 313: After recompilation print a gigantic (multi line) message to inform that one needs to reinstall packages (and add the instruction to do it, `pip install ...`).

bigflow/resources.py
```diff
@@ -174,9 +166,13 @@ def create_setup_body(project_name: str) -> str:
 '''
 
 
+@deprecated(
+    reason="""
+        This method doesn't work when called from pip-installed package.
+        Use `bigflow.build.reflect.materialize_setuppy` instead"""
```
beczkowb, line 172: I think we can automatically set the beam `options.setup_file = ...` option inside the BeamJob (of course in a separate issue/pull request, in the next release).

docs/technologies.md
```diff
@@ -61,20 +59,6 @@ The following line is the key:
 options.view_as(SetupOptions).setup_file = str(setup_file_path)
 ```
 
-If you want to provide requirements for your Beam process, you can do it through the
```
beczkowb, line 64: I think you can mention that we don't recommend using `requirements_file` because all the dependencies should be placed inside the setup.py, through resources/requirements.txt.

bigflow/build/__init__.py
```diff
@@ -0,0 +1,20 @@
+__all__ = []
+
+def setup(**kwargs):
+    """Speficy parameters of Bigflow project. Must me called from `setup.py`
```
iwko, line 4: """Specify parameters of Bigflow project. Must be called from `setup.py`

bigflow/build/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+    """
+    # Laziliy import 'bigflow.dist' as it depends on 'setuptools'
+    import bigflow.build.dist as d
+    return d.setup(**kwargs)
+
+
+# TODO: Remove in v2.0. COMPAT
```
iwko, line 11: ?

bigflow/build/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+import bigflow.build.dist as _dist
+from deprecated import deprecated
+for n in [
```
iwko, line 14: Why we use deprecated method here? Why not directly on listed objects?

bigflow/build/dev.py
```diff
@@ -0,0 +1,98 @@
...
+from typing import Optional, List, Union
+from pathlib import Path
+
+import bigflow.commons as bfc
+import bigflow.build.reflect
+
+
+logger = logging.getLogger(__name__)
+
+# Used by 'dist.py' to dump `bigflow.setup(...)` arguments.
```
iwko, line 23: This comment is redundant

bigflow/build/dev.py
```diff
@@ -0,0 +1,98 @@
...
+DUMP_PARAMS_SETUPPY_CMDARG = "__bigflow_dump_params"
+
+
+def read_setuppy_args(path_to_setup: Union[Path, str, None] = None) -> dict:
+    """Loads `setup.py`, returns all parameters of `bigflow.build.setup()` function.
+
+    This function doesn't unpack 'embeeded sdist' archive when package is installed via pip.
+    You could use `bigflow.build.materialize_setuppy` for such purposes, altough it is not recommended"""
```
iwko, line 31: You could use `bigflow.build.materialize_setuppy` for such purposes, **although** it is not recommended"""

bigflow/build/dev.py
```diff
@@ -0,0 +1,98 @@
...
+        logging.error(
+            "Project name mismatch: setup.PROJECT_NAME == %r, "
+            "but setup(name=%r). It is recommended to remove 'PROJECT_NAME' variable from 'project_setup.py'",
+            legacy_project_name, params.get('name'))
+
+    return params
+
+
+def _read_project_name_from_setup_legacy(direcotry: Path) -> Optional[str]:
+    # TODO: Please, remove me!
```
iwko, line 50: ?

bigflow/build/dev.py
```diff
@@ -0,0 +1,98 @@
...
+    with mock.patch('bigflow.build.dist.setup', lambda **kwargs: None):
+        sys_path_original = list(sys.path)
+        try:
+            sys.path.insert(0, direcotry)
+            import project_setup
+            return project_setup.PROJECT_NAME
+        except Exception:
+            return None
+        finally:
+            # del sys.modules['project_setup']
```
iwko, line 61: ?

bigflow/build/dist.py
```diff
@@ -85,27 +78,67 @@ def build_image(
...
+        bfc.remove_docker_image_from_local_registry(tag)
+
+
+def _hook_pregenerate_sdist(command_cls):
+    """
+    Wraps existing distutils.Command class.
+    Runs 'sdist' and copy resutl into 'build/bf-project.tar.gz'
+    """
+
+    class command(command_cls, distutils.cmd.Command):
```
iwko, line 96: UpperCase?

iwko:
In general too much comments for me. Also this PR is enormous and very hard to review. We should avoid changes of this size in single PR in the future

bigflow/build/dev.py
anjensan: ?

bigflow/build/__init__.py
anjensan: ?

bigflow/build/__init__.py
anjensan: We are still using those functions internally (inside `dist.py`), that is why they are not deprecated with decorator.

bigflow/build/dev.py
iwko: This comment does not provide any information.

bigflow/build/__init__.py
iwko: What info is provided here?

bigflow/build/__init__.py
iwko: Ok

iwko:
I will review this PR when all tests will be provided

bigflow/build/__init__.py
anjensan: This is a marker, which indicates "Someone need to remove this code after v2.0" 
I've changed comment text.

bigflow/build/dev.py
anjensan: Thanks, but functions are slight different - here we are using '$HOME' as top directory. Also I don't add extra dependency to 'resources' (but eventually we can move this into `commons`).

bigflow/build/dist.py
anjensan: Done.

bigflow/build/dev.py
anjensan: Agree, this is not a good comment. Fixed.

bigflow/build/dev.py
anjensan: Removed.

bigflow/build/dist.py
anjensan: I've refactored the code. Now it generate class with exactly same name.

docs/technologies.md
anjensan: I made this feature "hidden" for purpose.
If package added via 'beam..requirements_file` but not mentioned in project `requirements.in` it will be available only in production, making local development and writing tests more complicated...  For inexperienced users it might complicated to understand / debug.
Also advanced users might use this `requirements_file` anyway, even without mentioning in documentation (because they have 'beam' own docs).

bigflow/build/dist.py
anjensan: Done.

bigflow/build/dev.py
anjensan: Comment has been [changed].(https://github.com/allegro/bigflow/pull/164/commits/19a7807cfdc45e26923b8ae873d28c33f70a2ff4)

beczkowb (approved):

beczkowb:
The upgrade logic should erase `setup.py` from `.gitignore`. It's easy to miss. 

Also, check if the `.gitignore` in scaffolds (both open-source and internal.

beczkowb (changes requested):
check the comments

beczkowb:
The `bf build-requirements` command compiles only `requirements.in`. In our internal project, we have a requirements file tree, so the command should also compile child requirements.

beczkowb:
I know it's not a bug but looks like one

Zrzut ekranu 2020-12-11 o 12 46 29

anjensan:
@beczkowb 
During mitration v1 => 1.1 file ".gitignore" is patched (remove 'setup.py')
Also added check for 'bigflow' version inside 'pyproject.toml'.

anjensan:
@iwko All have been added.

beczkowb (approved):

Issue #175, opened 2020-12-09
Broken retry_count and delay


beczkowb:
The new `bigflow.Job` interface uses `retries` and `retries_delay` variables and the dag builder expects `retry_count` and `retry_pause_sec`.

PR #176, opened 2020-12-14
#175 fixing retry_count and retry_pause_sec

beczkowb:


bigflow/dagbuilder.py
```diff
@@ -88,7 +88,7 @@ def build_dag_operator(workflow_job, dependencies):
           root_folder=root_package_name,
           retries=job.retry_count if hasattr(job, 'retry_count') else 3,
           retry_delay=job.retry_pause_sec if hasattr(job, 'retry_pause_sec') else 60,
-          execution_timeout=job.execution_timeout))
+          execution_timeout=getattr(job, 'execution_timeout', DEFAULT_EXECUTION_TIMEOUT_IN_SECONDS)))
```
iwko, line 91: Maybe a better option will be to add the method get_execution_timeout to Job and move this logic there?

beczkowb: Not every job inherits from the Job interface (for example our jobs), so it can break the users' code.

iwko (approved):

PR #188, opened 2021-01-05
Add default src files via custom 'sdist' command, drop MAINFEST.in

anjensan:


beczkowb:
I think you should also update the scaffold

beczkowb (changes requested):
done

anjensan:
Scaffold is already updated, new projects will not get this 'MANIFEST.in'
However user still may create its own version of this file.

PR #199, opened 2021-01-13
E2e testing tutorial

beczkowb:


docs/e2e_testing.md
```diff
@@ -0,0 +1,398 @@
...
+
+* Daily transaction count
+* Daily transaction fee sum 
+
+The workflow saves the result in a BigQuery table.
+
+You will write tests for two implementations of that workflow. The first implementation 
+utilizes BigQuery (as the input and the output) and Dataflow (for processing).
+
+The second implementation utilizes only BigQuery, for both io and processing.
```
iwko, line 19: IO?

docs/e2e_testing.md
```diff
@@ -0,0 +1,398 @@
...
+
+```shell script
+cd ~/bigflow_cookbook/btc_aggregates_project
+rm -rf btc_aggregates/internationalports
+rm -rf btc_aggregates/wordcount
+```
+
+## Testing Dataflow + BigQuery implementation
+
+Take a look at the workflow implementation below. The important part (in the context of e2e testing) of that workflow is the configuration. Also, save the following code as a module inside the generated project: `btc_aggregates/btc_aggregates_df_bq.py`.
```
iwko, line 61: This code should be generated by bigflow start-project and should not require creation bt user

docs/e2e_testing.md
```diff
@@ -0,0 +1,398 @@
...
+
+    options.view_as(pipeline_options.WorkerOptions).machine_type = 'n2-standard-2'
+    options.view_as(pipeline_options.WorkerOptions).max_num_workers = 2
+    options.view_as(pipeline_options.StandardOptions).runner = conf['runner']
+    options.view_as(pipeline_options.SetupOptions).setup_file = str(materialize_setuppy().absolute())
+
+    return beam.Pipeline(options=options)
+
+
+class Transaction(NamedTuple):
```
iwko, line 128: @dataclass?

docs/e2e_testing.md
```diff
@@ -0,0 +1,398 @@
...
+
+    def to_dict(self):
+        return {
+            'fee_sum': self.fee_sum,
+            'count': self.count
+        }
+
+
+def month(runtime: str) -> str:
+    return runtime[:7] + '-01'
```
iwko, line 144: ?

docs/e2e_testing.md
```diff
@@ -0,0 +1,398 @@
...
+    BTC_TRANSACTIONS_TABLE_NAME)
+
+DATE = slice(0, 10)
+NOW = "2020-12-01 00:00:00"
+NOW_DT = datetime.fromisoformat(NOW + "+00:00")
+NOW_MINUS_ONE_MONTH = (NOW_DT - timedelta(weeks=4)).isoformat()[DATE]
+NOW_PLUS_ONE_MONTH = (NOW_DT + timedelta(weeks=4)).isoformat()[DATE]
+
+
+class BitcoinAggregatesWorkflowTestCase(SpawnIsolateMixin, unittest.TestCase):
```
iwko, line 248: class BitcoinAggregatesWorkflowTestCase(SpawnIsolateMixin):

docs/e2e_testing.md
beczkowb: I don't want to put it into the scaffold to avoid additional maintenance.

docs/e2e_testing.md
beczkowb: Mixin by definition shouldn't be a base class.

docs/e2e_testing.md
beczkowb: done

docs/e2e_testing.md
iwko: Right, so maybe we could create `class EndToEndTest(SpawnIsolateMixin, unittest.TestCase)` and then use is in tests?I think it will be cleaner solution

docs/e2e_testing.md
iwko: Putting additional code in docs also puts some additional maintenance ;) And because it is out of the main code it is harder to maintain imo. So maybe we should unify existing scaffold with e2e example?

docs/e2e_testing.md
beczkowb: done

docs/e2e_testing.md
beczkowb: done

docs/e2e_testing.md
beczkowb: Do you mean adding the `EndToEndTest` class to the bigflow codebase?

docs/e2e_testing.md
beczkowb: I have added the following note:

This tutorial uses `bigflow==1.1.3`

Now it requires pretty much zero maintenance. And it's still fine, even if there is a new bigflow version, like `1.6.8`, that tutorial still provides value because it shows what's the general idea behind e2e testing in bigflow. And at the same time, it's not a critical part of bigflow.

The only reasons I can see for adding this example to the scaffold are:

- We will avoid some incompatibilities with newer versions (though that tutorial should be ok with any version below 2.0).
- It's a bit easier to just generate the project and have the example inside the generated project.

The scaffold is a way to start a project, and not a place to store code for the tutorials.

@anjensan give us your thoughts, please

docs/e2e_testing.md
iwko: Yes

beczkowb: Then add it as an issue, I don't want to go through the release cycle for such a small change.

iwko: Ok

docs/e2e_testing.md
anjensan: datetime.fromisoformat(runtime).date().isoformat()

docs/e2e_testing.md
```diff
@@ -0,0 +1,412 @@
...
+
+    def extract_output(self, accumulator, *args, **kwargs) -> TransactionsAggregate:
+        return TransactionsAggregate(*accumulator)
+
+
+class CalculateBitcoinAggregatesJob(bf.Job):
+    def execute(self, context: bf.JobContext):
+        p = get_dataflow_pipeline(pipeline_config)
+        aggregates_table_id = f'{dataset_config["project_id"]}:{dataset_config["dataset_name"]}.{BTC_AGGREGATES_TABLE_NAME}'
+        runtime_month = month(context.runtime_str)
```
anjensan, line 211: context.runtime.date().isoformat()

docs/e2e_testing.md
```diff
@@ -0,0 +1,412 @@
...
+from datetime import datetime, timedelta
+
+from bigflow.testing import SpawnIsolateMixin
+
+from btc_aggregates.btc_aggregates_df_bq import (
+    btc_aggregates_workflow,
+    dataset as btc_aggregates_dataset,
+    BTC_TRANSACTIONS_TABLE_NAME)
+
+DATE = slice(0, 10)
```
anjensan, line 248: Using `slice` might be too complicated for someone who is not great in `python`.

docs/e2e_testing.md
```diff
@@ -0,0 +1,412 @@
...
+NOW = "2020-12-01 00:00:00"
+NOW_DT = datetime.fromisoformat(NOW + "+00:00")
+NOW_MINUS_ONE_MONTH = (NOW_DT - timedelta(weeks=4)).isoformat()[DATE]
```
anjensan, line 251: It is bad to manipulate dates this way (as a string). We can use it internally, but it shouldn't be a part of our "recommendations"

docs/e2e_testing.md
```diff
@@ -0,0 +1,412 @@
...
+        self.assertEqual(result, [{'fee_sum': 0.0, 'count': 0}])
+
+
+if __name__ == '__main__':
+    unittest.main()
+```
+
+Each of the two tests has the following schema:
+
+1. Preparing fake bitcoin transactions table and inserting test records.
```
anjensan, line 305: nit: We can add comments to the source code. It might help reader to understand where actually (what line) we a doing some thing.

anjensan (commented):
Some minor comments from me.

Also this tutorial contains huge pieces of code, which is not directly related to the topic.

IMO it is better to "define how to add tests" to some existing project: something from scaffold `new-project`, `wordcount` or so. Also such project can be just stored in subdirectory in git.

Applying changes to existing project will show to the user what actual steps should be 
 done and what pieces of source code added to project, rather than saying "just copy-paste this and you are done".  As for user it may be hard to draw a line "it is for e2e, and this is actual pipeline". At least drawing such line will take some time.

In general I'm ok with this tutorial. IMO it can be improved, but I'm not sure that it is really worth for doing right now.

docs/e2e_testing.md
beczkowb: The continuation of that discussion will be in Andreis comment.

docs/e2e_testing.md
beczkowb: done

docs/e2e_testing.md
beczkowb: done

docs/e2e_testing.md
beczkowb: done

docs/e2e_testing.md
beczkowb: done

docs/e2e_testing.md
beczkowb: done

docs/e2e_testing.md
```diff
@@ -0,0 +1,412 @@
...
+utilizes BigQuery (as the input and the output) and Dataflow (for processing).
+
+The second implementation utilizes only BigQuery, for both IO and processing.
+
+Workflows using BigQuery are interesting to test because there is no way emulate 
+BigQuery on a local machine.
+
+## Preparation
+
+For this tutorial, you need a fresh BigFlow project.
```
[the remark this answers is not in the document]
beczkowb: done (I think...)

beczkowb:
@anjensan @iwko About putting the example workflows into the scaffold - I think that we can throw away "old" bigquery and dataflow workflows from the scaffold and put these instead, but not in this PR. I will create an issue for that.

anjensan (approved):

PR #208, opened 2021-01-22
Refactoring and small improvements

anjensan:
PR #209 depends on this PR.


iwko:
Which task is connected to this pull-request?

anjensan:
> Which task is connected to this pull-request?

This is an outcome of "[opportunistic refactoring](https://martinfowler.com/bliki/OpportunisticRefactoring.html)" from #209

bigflow/build/pip.py
```diff
@@ -79,6 +100,25 @@ def maybe_recompile_requirements_file(req_txt: Path) -> bool:
...
+    logger.debug("Scan all req.in-like files: %s", req_in)
+    c = req_in.read_text()
+    yield c
+    for include in re.findall(r"\s*-r\s+(.*)", c):
+        include: str
+        fn = include.split("#", 1)[0].strip()
+        yield from _collect_all_input_files_content(req_in.parent / fn)
+
+
+def reqin_file_hash(req_in: Path):
```
beczkowb, line 113: use full names everywhere please, f.e. reqin -> requirements_in

bigflow/commons.py
```diff
@@ -27,34 +28,78 @@ def now(template: str = "%Y-%m-%d %H:00:00"):
     return datetime.now().strftime(template)
 
 
+class _OutDumper(threading.Thread):
```
beczkowb, line 31: Maybe `_StreamOutputDumper`?

test/mixins.py
```diff
@@ -13,7 +13,39 @@
 from pathlib import Path
 
 
-class TempCwdMixin(unittest.TestCase):
+class Mixin(unittest.TestCase):
```
beczkowb, line 16: Do we need this class? Can't you just do `FileUtilsMixin(unittest.TestCase)`?

beczkowb (changes requested):
needs work

bigflow/build/pip.py
anjensan: done

bigflow/commons.py
anjensan: done

test/mixins.py
anjensan: Well, strictly it is not required. We even can drop inheritance from `unittest.TestCase` completely (but user/client need to remember to put `TestCase` as the last class in the inheritance list). New class `Mixin` is added as "self-documented" code (and also eventually some mro-check may be added here).

beczkowb (approved):

PR #209, opened 2021-01-22
Detect/fix conflicts with preinstalled dependencies on dataflow workers

anjensan:
Added script to scrape Dataflow docs and generate list of preinstalled dependencies for various combinations of beam/python version.
Verify that used libraries matches to preinstalled versions for specified beam/python.
Add new command `bigflow codegen pin-dataflow-requirements` to generate `pinfile`: pin all used libs to version, used by 'dataflow'.


bigflow/build/dataflow/depscheck.py
```diff
@@ -0,0 +1,116 @@
+import re
```
iwko, line 1: Maybe `dependency_checker.py`?

iwko:
So now we would only support apache beam in versions 2.25, 2.26 and 2.27?

anjensan:
> So now we would only support apache beam in versions 2.25, 2.26 and 2.27?

Also 2.24 (I've added missing txt files). Maybe 2.28 if it will be released before bf 1.2
BTW it is more, than just 2.24 supported by bigflow 1.0

bigflow/build/dataflow/workerdeps/beam2.24_py3.5.txt
```diff
@@ -0,0 +1,111 @@
+# list of preinstalled dependencies on Dataflow workers
```
beczkowb, line 1: Is it possible that one of the dependencies gets a patch that list will be obsolete (so the user will constantly not match these requirements) ?

bigflow/build/dataflow/depscheck.py
```diff
@@ -0,0 +1,116 @@
...
+                ========================================
+            """),
+            "\n".join(
+                f" - {dep}: used {our_version}, beam has {worker_version}"
+                for dep, (our_version, worker_version) in conflicts.items()
+            ),
+        )
+
+
+def generate_dataflow_pins_file(req_path=None):
```
beczkowb, line 107: Im not sure if pins is the best name. Maybe overrides, dependency_fixes? Or something that describes the purpose.

bigflow/build/pip.py
```diff
@@ -169,3 +169,71 @@ def read_requirements(requirements_path: Path, recompile_check=True) -> List[str
     return result
 
 
+def generate_pinfile(
```
beczkowb, line 172: This function is long and not very descriptive. There are AT LEAST 2 parts worth implementing as separate functions:

* find conflicts
* resolve conflicts

bigflow/cli.py
```diff
@@ -812,5 +831,7 @@ def cli(raw_args) -> None:
         cli_logs(root_package)
     elif operation == 'build-requirements':
         _cli_build_requirements(parsed_args)
+    elif operation == 'codegen':
```
beczkowb, line 834: Maybe something more problem-related, like `bf fix-dataflow-dependency-conflicts`

test/build/dataflow/test_depscheck.py
```diff
@@ -0,0 +1,99 @@
...
+import bigflow.build.pip as bf_pip
+import bigflow.build.dataflow.depscheck as dc
+
+
+class DataflowDepsCheckTestCase(
+    mixins.TempCwdMixin,
+    mixins.FileUtilsMixin,
+    unittest.TestCase,
+):
+    def setUp(self):
```
beczkowb, line 17: It would be nice to have a E2E based on the CLI and a real project, what do you think?

beczkowb (changes requested):
needs work

bigflow/build/dataflow/depscheck.py
anjensan: IMO it is a too long name for python module.

beczkowb: `dependency_checker.py` is ok in my opinion

bigflow/build/dataflow/workerdeps/beam2.24_py3.5.txt
anjensan: I wasn't able to find a direct answer in the documentation.
However old "beam" versions (like 2.25) still uses "old" libraries (even don't get minor updates).
So IMO it is safe to assume that "versions are newer change". If they are - we will recognise this via 'git diff' after rerunning `grab_beam_worker_deps.py`.

bigflow/build/dataflow/depscheck.py
anjensan: AFAIK "pin" is a commonly used term. 
I found many links/documents which use "pinning dependencies" and/or "pin" for such meaning.
like https://before-you-ship.18f.gov/infrastructure/pinning-dependencies/

beczkowb: What I wanted to say is that "pins" describe the solution to the problem of fixing dependencies. So the question is – should that function tell what problem it solves or how it solves the problem (open question)?

test/build/dataflow/test_depscheck.py
anjensan: IMO it doesn't make a big sense - it interacts only with `pip-compile` tool.
And rest of `bigflow` knows nothing about such pinning (including `bigflow.dataflow` package).

bigflow/build/pip.py
anjensan: I'm not sure that I understand what you mean by saying "find conflicts" and "resolve conflicts".

This functions is *not* trying to find any conflicts - it receives "find conflicts" as a callback (3rd argument). This is needed because this code must be executed with empty "pinfile" and freshly recompiled "requirements.txt".
Such logic (actually 5 lines of linear code) may be extracted into something like `execute_with_empty_pinfile` (and callback converted to just a list of pins/strings), but I don't think it makes much sense. Did you mean that kind of refactoring?

I've moved `for` loop into a separate function `_try_incrementally_add_pins` -- do you mean this part while saying "resolve conflicts" ? 

Also I've extracted `_include_pinsfile_into_requirements` into a separate function.

bigflow/build/dataflow/workerdeps/grab_beam_worker_deps.py
```diff
@@ -0,0 +1,86 @@
...
+            yield {
+                'python_version': python_ver,
+                'beam_version': beam_ver,
+                'packages': packages,
+            }
+
+
+def main():
+
+    wdeps = list(grab_sdk_worker_dependencies())
```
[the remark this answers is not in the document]
anjensan: done

test/build/dataflow/test_depscheck.py
beczkowb: ok

bigflow/build/pip.py
beczkowb: "I've moved for loop into a separate function _try_incrementally_add_pins -- do you mean this part while saying "resolve conflicts" ?" – yes

bigflow/cli.py
anjensan: I don't very like to "litter" `bigflow` cmd-line interface with such "very occasional and specific" functions. One indicator  why it is "specific" case - very long and "ugly" name (either current `bf codegen pin-dataflow-requirements` or proposed `bf fix-dataflow-dependency-conflicts`).

Idea was to add a new "umbrella" command for such tools - eventually we might want to add something like `upgrade-docker-file-with-new-python`, `create-pyproject-toml`, `pin-dataproc-requirements`, `migrate-to-bigflow-1.2` etc etc ...)

I'm not insist on this idea - we can just put all such things into a global namespace if you would prefer.

Regarding actual command name (`pin-dataflow-requirements` vs `fix-dataflow-dependency-conflicts`).
IMO this is not very important, as this command is printed to user as a warning "please run this xxx to fix an issue".

beczkowb: Idea was to add a new "umbrella" command for such tools - eventually we might want to add something like upgrade-docker-file-with-new-python, create-pyproject-toml, pin-dataproc-requirements, migrate-to-bigflow-1.2 etc etc ...)

ok

bigflow/build/dataflow/depscheck.py
anjensan: Ok, if you both think it should be `dependency_checker`. I've renamed this module.

bigflow/build/dataflow/depscheck.py
anjensan: If you about function name - yes, it might be something like `sync_requirements_with_dataflow_workers`, `fix_requirements_with_dataflow_workers` etc etc. Do you have something in mind?

iwko:
>So now we would only support apache beam in versions 2.25, 2.26 and 2.27?

>Also 2.24 (I've added missing txt files). Maybe 2.28 if it will be released before bf 1.2
>BTW it is more, than just 2.24 supported by bigflow 1.0

Can't we just provide bigflow with one beam version and support just that one version ignoring what beam version user have installed?

bigflow/build/dataflow/depscheck.py
beczkowb: `sync_requirements_with_dataflow_workers ` is fine

anjensan:
> Can't we just provide bigflow with one beam version and support just that one version ignoring what beam version user have installed?

It was my first idea to address this issue (or as an alternative to create a separate "extras" for all supported versions).

But list of preinstalled dependencies is quite big (around 100 libraries), and pip doesn't provide any way to restrict library version, but not actually require it (or I don't know about such mechanism). So the only option - just add all libs as unconditional dependencies. Looks like not a good idea.

Also such approach doesn't allow user to "upgrade" any of the libraries. For example

iwko:
👍

iwko (approved):

Issue #194, opened 2021-01-08
Non-deterministic get_version function


beczkowb:
Currently, the `bigflow.version.get_version` function adds a random suffix to the version if the repository is dirty. There are 2 problems with that fact:

- A developer has to use the function only once, otherwise, a developer can get 2 different versions
- When a user runs `build-image/dags/package` commands separately, each command might generate a different version

We need to make the suffix generator deterministic.

anjensan:
Link #154

Issue #135, opened 2020-11-05
--no-snapshot flag for `bf build`


beczkowb:
Currently, the versioning system adds a unique id if a repository is dirty (so you don't have to commit every time you make a local change and build). In CI/CD servers it would be useful to turn off this behavior, so you don't have to add things (if they occur on build) to gitignore.

The `--no-snapshot` would prevent from adding a unique id on build:

`bf build --no-snapshot`

PR #226, opened 2021-02-25
Change git-based project version detection

### LINKED ISSUE #154: Use PEP 440 compliant version for snapthots
During build on dirty git repo:

`dist.py:476: UserWarning: The version specified ('0.1.0SNAPSHOT6d2bde52') is an invalid version, this may not work as expected with newer versions of setuptools, pip, and PyPI. Please see PEP 440 for more details.`

anjensan:
Now function `bigflow.version.get_version` uses `git describe` to find nearest tag. 
When there is such tag - it is used as a version.
If there are some extra commits on top of this tag, then version is constructed as:
{version}.dev{distance-to-nearest-tag}+g{commit hash}{.dirty}
for example: 1.2.dev34+g25f38a38.dirty

No "0" patch version is added - in practice all python tools (pip etc) assume `1.2 == 1.2.0`

Generated versions are PEP440 compliant and normalised (fix #154, no annoying warnings)
Also version does not include any random number / timestamp / etc, it solves #194 
Untracked (new) files do not trigger appending of ".dirty" (corresponds to default git behaviour), it solves #135


beczkowb (changes requested):
Make sure that you can override images (so the working on a dirty repo is still ok). Also, update the documentation.

anjensan:
Changed "dirty" versions: instead of suffix `.dirty` now version ends with tree-hash (included all changes from current git index + all tracked modified files added by `git add -u`), eg: `1.23.dev45+g592febad.d4fb90c1c`
Untracked files are still ignored.

Note: generated hash is a real tree-ish, which implicitly added to .git repo.
It gives nice ability to recover files (git ls-tree, git diff etc, git checkout) having only version id (when version was generated on the same git repo).

beczkowb (approved):
LGTM

PR #233, opened 2021-03-20
changing RuntimeError to TimeoutError in BeamJob timeout exception case

beczkowb:


iwko (approved):

bigflow/dataflow.py
```diff
@@ -54,7 +54,7 @@ def execute(self, context: JobContext):
             result.wait_until_finish(timeout_in_milliseconds)
             if not result.is_in_terminal_state():
                 result.cancel()
-                raise RuntimeError(f'Job {self.id} timed out ({self.execution_timeout_sec})')
+                raise TimeoutError(f'Job {self.id} timed out after {self.execution_timeout_sec} seconds')
```
anjensan, line 57: AFAIK TimoutError is a subclass of `OSError`, which has attributes `errno`, `filename` etc.
Maybe it is not a best choice for this place.

anjensan (approved):
approve. but maybe it is better to create custom exception

beczkowb:
not needed

PR #234, opened 2021-03-23
Infra generator

beczkowb:


bigflow/scaffold/infra.py
```diff
@@ -0,0 +1,159 @@
...
+
+
+def cloud_composer(
+        gcp_project_id: str,
+        bigflow_project_name: str,
+        environment_name: str = 'dev',
+        region: str = 'europe-west1',
+        zone: str = 'europe-west1-d') -> tp.Tuple[str, InfrastructureDestroyer, InfrastructureCreator]:
+    composer_name = f'{BIGFLOW_INFRA_PREFIX}-composer-{bigflow_project_name}-{environment_name}-{region}'
+    cloud_nat_name, cloud_nat_destroyer, cloud_nat_creator = _create_cloud_nat(
```
iwko, line 128: Should not be this call made in create method?

bigflow/scaffold/infra.py
```diff
@@ -0,0 +1,159 @@
...
+        gcp_project_id, bigflow_project_name, environment_name, region)
+
+    @secure_destroy(f'{composer_name} (Google Cloud Composer)')
+    def destroy() -> None:
```
iwko, line 132: Shouldn't it call also destroy methods of nat_creator and router?

anjensan:
Very nice idea, I like.

Would recommend to slightly refactor this by replacing `tp.Tuple[str, InfrastructureDestroyer, InfrastructureCreator]` with a base class. It might be easier to extend such solution (add retrying etc), and also it might be easier to read by someone who is not experienced with a FP style. Something like:

```python
class ResourceManager:

    name: str
    depends: Iterable[ResourceManager] = ()

    def destroy(self): 
        raise NotImplementedError
    
    def create(self): 
        raise NotImplementedError


def _with_all_subresources(r: ResourceManager):   # plain list of res
    for s in r.depends:
        yield from _with_all_subresources(s)
    yield r

    
def try_create(res):  # also may be moved to method inside baseclass
    logger.info("Trying to create %s", self.name)
    maybe_leftover = []

    try:
        for s in _with_all_subresources(res):
            maybe_leftover.append(s)
            s.create()
    except Exception:
        logger.exception("OMG :(")
    else:
        logger.info("ok")
        return

    logger.info("Cleanup leftovers")
    for s in reversed(maybe_leftover):
        s.destroy()
```

As most of potential resources are "gcloud" objects, also may be added such mixin/baseclass:
```python
class GCloudCmdResourceManager(ResourceManager):

    create_cmd: List[str]
    destroy_cmd: List[str]

    def create(self):
        run_process(self.create_cmd)
    
    def destroy(self):
        run_process(self.destroy_cmd)
```

Then resource creation might look like:
```python

class cloud_composer(GCloudCmdResourceManager):
    
    def __init__(self, gcp_project_id, bigflow_project_name, env, region, zone):

        self.name = f'{BIGFLOW_INFRA_PREFIX}-composer-{bigflow_project_name}-{environment_name}-{region}'

        self.depends = [
            _create_cloud_nat(gcp_project_id, bigflow_project_name, environment_name, region),
        ]

        self.create_cmd = [
            'gcloud', 'composer',
            'environments', 'create', self.name,
            ...
        ]

        self.destroy_cmd = [
            'gcloud', 'composer',
            'environments', 'delete', self.name,
            '--location', region,
            '--project', gcp_project_id,
            '--quiet',
        ]
```

bigflow/scaffold/infra.py
```diff
@@ -0,0 +1,159 @@
...
+def _create_cloud_router(
+        gcp_project_id: str,
+        bigflow_project_name: str,
+        environment_name: str,
+        region: str) -> tp.Tuple[str, InfrastructureDestroyer, InfrastructureCreator]:
+    router_name = f'{BIGFLOW_INFRA_PREFIX}-router-{bigflow_project_name}-{environment_name}-{region}'
+
+    @secure_destroy(f'{router_name} (Google Cloud router)')
+    def destroy() -> None:
+        logger.info(run_process([
```
anjensan, line 39: BTW `run_process` may dump all its output to console with `verbose=True` (it is a default).

bigflow/scaffold/infra.py
```diff
@@ -0,0 +1,159 @@
...
+        ]))
+
+    return router_name, destroy, create
+
+
+def _create_cloud_nat(
+        gcp_project_id: str,
+        bigflow_project_name: str,
+        environment_name: str,
+        region: str) -> tp.Tuple[str, InfrastructureDestroyer, InfrastructureCreator]:
```
anjensan, line 63: A good candidate for type-alias.

bigflow/scaffold/infra.py
```diff
@@ -0,0 +1,159 @@
...
+    return composer_name, destroy, create
+
+
+def try_create(
+        name: str,
+        destroyer: InfrastructureDestroyer,
+        creator: InfrastructureCreator) -> None:
+    try:
+        creator()
+    except Exception as e:
```
anjensan, line 156:
```python
except Exception:
        logger.exception(f"Can't create {name}. Trying to destroy leftovers.")
        destroyer()
```

iwko: I would mix those. Logging real exception message is really helfpul.

bigflow/scaffold/infra.py
anjensan: fixup: `run_process` always dumps its output to `logger`. Flag `verbose` switches between `info` and `debug` log levels.

bigflow/scaffold/infra.py
anjensan: `logger.exception` prints both log message and exception information (traceback + exception text). [example](https://onecompiler.com/python/3wshm4c2b)
In contrast with Java there is no need to pass exception object into logger explicitly.
But you still could do it by adding named arg 'exc_info` [example](https://onecompiler.com/python/3wshkm2c4)

beczkowb: done

bigflow/scaffold/infra.py
beczkowb: done

bigflow/scaffold/infra.py
beczkowb: refactored

anjensan:
I suppose it may not work in such scenario:
 - successfully created nat via "gcloud compute routers nats create ..."
 - failed to run "gcloud compute composer create ..."
 - trying to delete composer via "gcloud compute composer delete ..."
 - exception => code inside _CloudNat is not invoked.
 
 Maybe explicitly ignore errors inside destroys.
 Or add `check=False` to `run_process`

test/test_scaffold.py
```diff
@@ -302,4 +301,129 @@ def test_conditional_render(self):
...
+
+    @mock.patch('bigflow.scaffold.infra.run_process')
+    @mock.patch('bigflow.scaffold.infra._composer_create_command')
+    def test_should_destroy_leftovers_when_error_occurs_during_creation(self, composer_create_command_mock, run_process_mock):
+        # given
+        def raise_error(*args):
+            raise ValueError('panic!')
+
+        gcp_project_id, bigflow_project_name, environment_name, region, zone = self.infrastructure_parameters
+        composer_create_command_mock.side_effect = raise_error
```
anjensan, line 375: composer delete command also may fail

beczkowb: invoking `secure_destroy` should handle the situation

anjensan (approved):

PR #235, opened 2021-03-23
Fix beam workerdeps detection, bump max version to 2.28

anjensan:
Now beam worker deps detection doesn't work with 'whl' package


iwko:
Are you sure that dataflow is working fine with apache beam 2.28? Yesterday I tried to run jobs with 2.28 and they were hanging forever

iwko (changes requested):
Needs to resolve beam 2.28 poblem

iwko (approved):

PR #239, opened 2021-04-13
Rework build system - embed setup properties to pyproject.toml

anjensan:
Massive update 'bigflow.build' packages

- remove deprecated/unused code -- build directories ('.image', '.dags', 'build', etc) are no longer configurable. Those settings is unusable in 1.1-1.2 anyways because in some places we had hardcoded paths.

- read project settings (project name, version, packages) from both 'pyproject.toml' AND 'setup.py'
  settings from setup.py has higher priority, settings from pyproject.toml used as defaults
  this allows to write "dynamic" settings (like "version") to pyproject.toml during compiling phase
  
- 'bigflow' tool don't call 'setup.py project_build', instead directly triggers distuitls internal mechanics. This results in slightly better logging (respect -v flag, etc), a bit more straightforward code. 

- side effect: 'setup.py' is optional. project building will work even without 'setup.py' (project name should be specified in pyproject.toml)

- all existing tests have been fixed, however new tests will be added soon


bigflow/build/dist.py
```diff
@@ -27,74 +21,31 @@
...
-    build_docker_image(project_dir, tag)
-    try:
-        export_docker_image_to_file(tag, image_dir, version)
-        shutil.copyfile(deployment_config, image_dir / deployment_config.name)
-    finally:
-        bf_commons.remove_docker_image_from_local_registry(tag)
+class BigflowDistribution(distutils.dist.Distribution):
+    """Customized Distribution for bigflow projects. Add custom commands, allow access to bigflow project spec."""
+
+    def __init__(self, attrs=None):
```
beczkowb, line 36: Type hints please, I'm wondering why there is `attrs = dict(attrs or {})` below

bigflow/build/dist.py
```diff
@@ -27,74 +21,31 @@
...
+        self.bigflow_project_spec = None
+        attrs = dict(attrs or {})
+        cmdclass = attrs.setdefault('cmdclass', {})
+        cmdclass['build_project'] = build_project
+        cmdclass['sdist'] = sdist
+        super().__init__(attrs)
+
+    def get_command_class(self, command):
```
beczkowb, line 44: If we want to use mypy soon, then I think we should use hints as much as possible

bigflow/build/dist.py
```diff
@@ -27,74 +21,31 @@
...
-    os.mkdir(image_dir)
-    tag = bf_commons.build_docker_image_tag(docker_repository, version)
-    build_docker_image(project_dir, tag)
-    try:
-        export_docker_image_to_file(tag, image_dir, version)
-        shutil.copyfile(deployment_config, image_dir / deployment_config.name)
-    finally:
-        bf_commons.remove_docker_image_from_local_registry(tag)
+class BigflowDistribution(distutils.dist.Distribution):
+    """Customized Distribution for bigflow projects. Add custom commands, allow access to bigflow project spec."""
```
iwko, line 34: What is project spec?

bigflow/build/operate.py
```diff
@@ -0,0 +1,147 @@
...
+import bigflow.build.dataflow.dependency_checker
+import bigflow.commons as bf_commons
+
+from bigflow.build.spec import BigflowProjectSpec
+
+
+logger = logging.getLogger(__name__)
+
+
+def run_tests(prj: BigflowProjectSpec):
```
iwko, line 27: Not used arg

bigflow/build/operate.py
```diff
@@ -0,0 +1,147 @@
...
+    logger.info('Runing tests...')
+    output_dir = "build/junit-reports"
+    try:
+        bf_commons.run_process([
+            "python", "-m", "xmlrunner", "discover",
+            "-s", ".",
+            # "-t", project_dir,
```
iwko, line 34: To remove?

bigflow/build/operate.py
```diff
@@ -0,0 +1,147 @@
...
+        dconf_file = Path(prj.deployment_config_file)
+        shutil.copyfile(dconf_file, image_dir / dconf_file.name)
+    finally:
+        bf_commons.remove_docker_image_from_local_registry(tag)
+
+    logger.info("Docker image was built")
+
+
+def build_dags(
+    prj: BigflowProjectSpec,
```
iwko, line 78: project_spec

bigflow/build/operate.py
```diff
@@ -0,0 +1,147 @@
...
+    start_time: str,
+    workflow_id: typing.Optional[str] = None,
+):
+    logger.info("Building airflow DAGs...")
+    clear_dags_leftovers(prj)
+
+    # FIXME: DON'T USE 'bigflow.cli'!
```
iwko, line 85: Why do not use bigflow.cli?

bigflow/build/operate.py
```diff
@@ -0,0 +1,147 @@
...
+    from bigflow.cli import _valid_datetime, walk_workflows
+    _valid_datetime(start_time)
+
+    cnt = 0
+    for root_package in prj.packages:
+        if "." in root_package:
+            # leaf package
```
iwko, line 92: redundant comment

bigflow/build/reflect.py
```diff
@@ -152,6 +171,27 @@ def _locate_plain_setuppy_on_fs(stack):
         return None
 
 
+@public()
+def get_project_spec(
+    project_name: Optional[str] = None,
+) -> BigflowProjectSpec:
+    """Read & parses project spec.
+    This function should be used from inside project.
```
iwko, line 179: Could you be more specific here? This is public method but docstring says `should be used from inside project.`

bigflow/build/spec.py
```diff
@@ -0,0 +1,262 @@
...
+
+import bigflow.commons as bf_commons
+
+
+logger = logging.getLogger(__name__)
+
+
+_ListStr = typing.List[str]
+
+# copied from `distutils.dist.DistributionMetadata`
```
iwko, line 30: So why we wont use `from distutils.dist import DistributionMetadata`?

bigflow/build/spec.py
```diff
@@ -0,0 +1,262 @@
...
+    # Bigflow-specific information and options
+    docker_repository: str
+    resources_dir: str
+    deployment_config_file: str
+    project_requirements_file: str
+
+    # Known package metainformation (author, url, description etc, see _PROJECT_METAINFO_KEYS)
+    metainfo: typing.Dict[str, typing.Any]
+
+    # Just bypass any unknow options to 'distutils'
```
iwko, line 67: Just bypass any unknow**n** options to 'distutils'

bigflow/build/spec.py
```diff
@@ -0,0 +1,262 @@
...
+def get_project_spec(dir: Path = None):
+    """Reads project spec from `setup.py` and/or `pyproject.toml`.
+    Memoize results (key = project path).
+    """
+
+    dir = dir or Path.cwd()
+    return read_project_spec(dir)
+
+
+def _mabye_read_pyproject(dir: Path):
```
iwko, line 160: _maybe_read_pyproject

bigflow/build/spec.py
```diff
@@ -0,0 +1,262 @@
+"""Read and parse bigflow project configuration (setup.py / pyproject.toml)"""
```
iwko, line 1: Missing typehints in this file

bigflow/build/spec.py
```diff
@@ -0,0 +1,262 @@
...
+    logger.info("Read project requirements from %s", project_requirements_file)
+    req_txt = Path(project_requirements_file)
+    recompiled = bigflow.build.pip.maybe_recompile_requirements_file(req_txt)
+    if recompiled:
+        logger.warning(textwrap.dedent(f"""
+            !!! Requirements file was recompiled, you need to reinstall packages.
+            !!! Run this command from your virtualenv:
+            pip install -r {req_txt}
+        """))
+    bigflow.build.dataflow.dependency_checker.check_beam_worker_dependencies_conflict(req_txt)  # XXX
```
iwko, line 228: redundant comment

bigflow/build/spec.py
```diff
@@ -0,0 +1,262 @@
...
+    if docker_repository is None:
+        raise ValueError(f"Can't find the 'docker_repository' property in the specified config file: {deployment_config_file}")
+    return docker_repository
+
+
+def secure_get_version() -> str:
+    logger.debug("Autodetected project version using git")
+    try:
+        version = bigflow.version.get_version()
+        logger.info("Autodected project version is %s", version)
```
iwko, line 257: Autodetected project version is

bigflow/cli.py
```diff
@@ -188,6 +189,7 @@ def find_root_package(project_name: Optional[str], project_dir: Optional[str]) -
     @return: Path
     """
     if project_name is not None:
+        project_name = project_name.replace("-", "_")
```
iwko, line 192: Why do we change this in other way here https://github.com/allegro/bigflow/blob/0e0415bdd6035753e4a7eb1f067fd15df894ae52/bigflow/cli.py#L192

bigflow/cli.py
```diff
@@ -543,63 +545,39 @@ def _cli_deploy_image(args):
 def find_image_file():
     # TODO parametrize ".image" using settings from build.py
     files = glob1(".image", "*-*.tar")
+    print("Z" * 100, list(files))
```
iwko, line 548: What is purpose of this print?

test/bf-projects/bf_simple_v10/project_setup.py
```diff
@@ -3,4 +3,4 @@
 PROJECT_NAME = "bf_simple_v10"
```
iwko, line 3: maybe change bf-projects directory to bf_projects?

test/mixins.py
```diff
@@ -209,3 +211,12 @@ def bigflow_run(self, cmd, **kwargs):
     def bigflow_spawn(self, cmd, **kwargs):
         cmd = ["python", "-m", "bigflow", *cmd]
         return self.subprocess_spawn(cmd, **kwargs)
+
+
+class ABCTestCase(unittest.TestCase):
+
+    __test__ = True
```
iwko, line 218: Why do we need this property?

bigflow/build/operate.py
anjensan: Because:
1 - circular dependency (cli -> build -> cli)
2 - cli-related stuff is sort of "UI", and should be separated from actual logic
3 - cli initialisation might take too long (init agrparser, autocomplete, load creds from ~/.blabla etc). Ideally it should not be involved into "python setup.py ..." process.

bigflow/build/spec.py
anjensan: Because such variable (list of strings) is not public (prefixed with "_").

bigflow/cli.py
anjensan: This line prints very nice and beautiful comment. :D
```
ZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZ
```

Ok, it was just a debug print :)

test/mixins.py
anjensan: It allows to disable unittest classes by adding '__test__ = False'
But tests are enabled by default.

bigflow/build/dist.py
anjensan: Because `attrs` might be None according to 'dist.Distribution' sources.

bigflow/build/dist.py
anjensan: IMO we need to put type-hits to all public (aka @public functions).
For internal stuff it much less important (especially for such cases, with weird integrations).

For example in our case `distutils` doesn't provide any typehints - to adding a few our custom doesn't change the game.

bigflow/build/operate.py
anjensan: Fixed.

bigflow/build/operate.py
anjensan: Maybe. For now I keep it.
Eventually I want to replace `python -m xmlrunner` with `pytest`.

bigflow/build/operate.py
anjensan: fixed

bigflow/build/operate.py
anjensan: totally agree

bigflow/build/reflect.py
anjensan: I've changed docstring.

bigflow/build/spec.py
anjensan: fixed

bigflow/build/spec.py
anjensan: fixed

bigflow/build/spec.py
anjensan: fixed

bigflow/build/spec.py
anjensan: not only redundant, but also a duplicate

bigflow/cli.py
anjensan: Python uses "-" in package names, but "_" in module / file names.

test/bf-projects/bf_simple_v10/project_setup.py
anjensan: This dir should be not "python-importable".
So I'd keep "-" to prevent this.

bigflow/build/reflect.py
iwko: 👍

beczkowb (approved):

Issue #238, opened 2021-04-09
BeamJob DirectRunner error


beczkowb:
The DirectRunner result doesn't have the `is_in_terminal_state` method. Because of that, the following code fails for the DirectRunner:

```       
 if self.wait_until_finish and self.execution_timeout_sec:
            timeout_in_milliseconds = 1000 * (self.execution_timeout_sec - self.pipeline_level_execution_timeout_shift)
            result.wait_until_finish(timeout_in_milliseconds)
            if not result.is_in_terminal_state():
                result.cancel()
                raise RuntimeError(f'Job {self.id} timed out ({self.execution_timeout_sec})')
```

beczkowb:
```
AttributeError: 'RunnerResult' object has no attribute 'is_in_terminal_state'
```

PR #248, opened 2021-04-20
Use custom docker image on Dataflow workers

anjensan:
Changes in `bigflow.dataflow` module:
 - BeamJob accepts 'dict' as 'pipeline_options' parameter
 - BeamJob automatically puts 'job_name' and 'setup_file' parameters (when they are not present in 'pipeline_options')
 - New parameter 'use_docker_image'. When it is a string - it is used as 'worker_harness_container_image` pipeline option. If it is a `True` - bigflow use the same docker image name as '.dags/' and 'bigflow deploy'.
 - Minor refactorign - easier to overload some parts in sublcasses (if user needs some specific tuning).
 - 'entry_point' might take any number of args/kwargs (same style as 'threading.Thread')

Missing features: 
 - No documentation updates (as it is an experimental feature).
 - Prechecking that image name is valid (or/and deployed).

### DIFF
--- a/bigflow/dataflow.py
+++ b/bigflow/dataflow.py
@@ -101,6 +101,9 @@ class BeamJob(Job):
                 drop_default=True,
                 retain_unknown_options=True,
             )
+            pipeline_options2 = PipelineOptions.from_dictionary(pipeline_options).get_all_options(
+                drop_default=True, retain_unknown_options=True)
+            assert pipeline_options == pipeline_options2, "During convertsion PipelineOptions<>dict some parameters was gone"
 
         self.pipeline_options = dict(pipeline_options or {})
         assert PipelineOptions.from_dictionary(self.pipeline_options)
--- a/bigflow/dataflow.py
+++ b/bigflow/dataflow.py
@@ -1,12 +1,18 @@
 import typing
 import logging
+import uuid
 
 from apache_beam import Pipeline
-from apache_beam.options.pipeline_options import PipelineOptions, GoogleCloudOptions
+from apache_beam.options.pipeline_options import PipelineOptions
+from apache_beam.runners.runner import PipelineResult, PipelineState
 
-from bigflow.commons import public
+from typing import Dict, Union
+
+from bigflow.commons import public, build_docker_image_tag
 from bigflow.workflow import Job, JobContext
-from bigflow.workflow import DEFAULT_EXECUTION_TIMEOUT_IN_SECONDS, DEFAULT_PIPELINE_LEVEL_EXECUTION_TIMEOUT_SHIFT_IN_SECONDS
+
+import bigflow.build.reflect
+
 
 logger = logging.getLogger(__file__)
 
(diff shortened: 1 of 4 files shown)


bigflow/dataflow.py
```diff
@@ -1,69 +1,204 @@
...
 from bigflow.workflow import Job, JobContext
-from bigflow.workflow import DEFAULT_EXECUTION_TIMEOUT_IN_SECONDS, DEFAULT_PIPELINE_LEVEL_EXECUTION_TIMEOUT_SHIFT_IN_SECONDS
+
+import bigflow.build.reflect
+
 
 logger = logging.getLogger(__file__)
 
 
+class PipelineOptionsDict(TypedDict, total=False):
```
beczkowb, line 25: I think that you can also add mostly pre-configured pipeline options class for our internal users:
```
runner = 'DataflowRunner'
experiments = ['shuffle_mode=service']
machine_type = 'n2-standard-4'
max_num_workers = 200
disk_size_gb = 100
autoscaling_algorithm = 'THROUGHPUT_BASED'
network = 'default'
use_public_ips = False
region = 'europe-west1'
staging_location and temp_location = it would be nice to automatically create these in the future
setup_file = str(reflect.materialize_setuppy())
job_name = somehow infered from the BeamJob?
```

bigflow/dataflow.py
```diff
@@ -1,69 +1,204 @@
...
+    disk_size_gb: int
+    flexrs_goal: str
+
+
 @public()
 class BeamJob(Job):
 
+    pipeline_level_execution_timeout_shift = 120  # 2 minutes
+
     def __init__(
```
beczkowb, line 67: Didn't you break the backward compatibility (by, for example, moving pipeline options) ?

bigflow/dataflow.py
```diff
@@ -1,69 +1,204 @@
...
+                imgid = self.use_docker_image
+            else:
+                logger.debug("Infer docker image name for current project")
+                pspec = bigflow.build.reflect.get_project_spec(self._project_path)
+                imgid = build_docker_image_tag(pspec.docker_repository, pspec.version)
+            logger.info("Use docker image %s for beam workers", imgid)
 
+            options['worker_harness_container_image'] = imgid
+            experiments = options.setdefault('experiments', [])
+            if 'use_runner_v2' not in experiments:
```
beczkowb, line 202: I'm sure if that should be a default, it's buggy

bigflow/dataflow.py
iwko, line 25: Shouldn't be this public? Also, what is total parameter?

bigflow/dataflow.py
iwko: +1

bigflow/dataflow.py
```diff
@@ -1,69 +1,204 @@
...
             wait_until_finish: bool = True,
-            execution_timeout_sec: int = DEFAULT_EXECUTION_TIMEOUT_IN_SECONDS,
             test_pipeline: Pipeline = None,
+            execution_timeout_sec: int = None,
+            use_docker_image: typing.Union[str, bool] = False,
+            project_name=None,
+            **kwargs,
     ):
-        if bool(test_pipeline) == bool(pipeline_options):
+        if bool(test_pipeline) == bool(pipeline_options is not None):
```
iwko, line 85: Why you've extended the condition?

bigflow/dataflow.py
```diff
@@ -1,69 +1,204 @@
...
             raise ValueError("One of the pipeline and pipeline_options must be provided.")
+
         if not wait_until_finish and execution_timeout_sec:
             raise ValueError("If wait_until_finish_set to False execution_timeout can not be used.")
 
-        self.id = id
-        self.entry_point = entry_point
-        self.entry_point_arguments = entry_point_arguments
-        self.pipeline_options = pipeline_options
+        super().__init__(
```
iwko, line 91: Do we need this? Outside the __init__?

bigflow/dataflow.py
```diff
@@ -1,69 +1,204 @@
...
+
+        logger.info("wait pipeline result...")
+        self.wait_pipeline_result(result)
+
+    def wait_pipeline_result(self, result: PipelineResult):
         if self.wait_until_finish and self.execution_timeout_sec:
             timeout_in_milliseconds = 1000 * (self.execution_timeout_sec - self.pipeline_level_execution_timeout_shift)
             result.wait_until_finish(timeout_in_milliseconds)
-            if not result.is_in_terminal_state():
+            if not PipelineState.is_terminal(result.state):
```
iwko, line 141: What's the difference?

test/test_dataflow.py
```diff
@@ -1,70 +1,69 @@
...
+            | 'FilterWords'       >> beam.Filter(lambda w: w in words_to_filter)
+            | 'MapToCount'        >> beam.Map(lambda w: (w, 1))
+            | 'GroupWords'        >> beam.GroupByKey()
+            | 'Count'             >> beam.MapTuple(self.count)
+        )
+
+    def run(self, pipeline: Pipeline, context: JobContext, driver_arguments: dict):
+        return self.run_ex(pipeline, context, driver_arguments['words_to_count'], driver_arguments['words_to_filter'])
+
+    def nope(self, pipeline, context):
```
iwko, line 48: nope?

test/test_dataflow.py
```diff
@@ -96,132 +130,126 @@ def test_should_run_beam_job(self, is_in_terminal_state_mock):
...
     @patch.object(RunnerResult, 'cancel')
     @patch.object(RunnerResult, 'wait_until_finish')
-    def test_should_run_beam_job_with_timeout_with_cancel(self, wait_until_finish_mock, cancel_mock,
-                                                          is_in_terminal_state_mock):
+    def test_should_run_beam_job_with_timeout_with_cancel(
+        self,
+        wait_until_finish_mock,
+        cancel_mock,
+        state_mock,
+        locate_project_mock,
```
iwko, line 141: Is this used?

test/test_dataflow.py
```diff
@@ -96,132 +130,126 @@ def test_should_run_beam_job(self, is_in_terminal_state_mock):
         # and sets default value for execution_timeout_sec
         self.assertEqual(job.execution_timeout_sec, DEFAULT_EXECUTION_TIMEOUT_IN_SECONDS)
 
-    @patch.object(RunnerResult, 'is_in_terminal_state', create=True)
+    @patch.object(RunnerResult, 'state', new_callable=mock.PropertyMock)
```
iwko, line 133: What was wrong with previous mock?

test/test_dataflow.py
```diff
@@ -96,132 +130,126 @@ def test_should_run_beam_job(self, is_in_terminal_state_mock):
...
+        job.execute(JobContext.make())
 
         # then
         self.assertEqual(cancel_mock.call_count, 0)
         wait_until_finish_mock.assert_called_with((600 - DEFAULT_PIPELINE_LEVEL_EXECUTION_TIMEOUT_SHIFT_IN_SECONDS) * 1000)
 
-    def test_should_throw_if_wait_until_finish_set_to_false_and_execution_timeout_passed(self):
+    def test_should_throw_if_wait_until_finish_set_to_false_and_execution_timeout_passed(
+        self,
+        locate_project_mock,
```
iwko, line 193: Is it used?

bigflow/dataflow.py
anjensan: Options 'setup_file' and 'job_name' are already provided in this PR (via method BeamJob.set_default_pipline_options).
I'll definitely add 'runner'...  Not sure about rest of parameters.

iwko:
Missing test for case when use_docker_image is string

bigflow/dataflow.py
anjensan: It marks all keys as optional, handled by TotalDict metaclass (`typing._TypedDictMeta.__new__`).
https://docs.python.org/3/library/typing.html#typing.TypedDict
> By default, all keys must be present in a TypedDict. It is possible to override this by specifying totality. Usage:

bigflow/dataflow.py
anjensan: This is a fix for #238

test/test_dataflow.py
anjensan: 'locate_project_mock' is not used, but mock is still passed as argument by `unitetest.mock`

test/test_dataflow.py
anjensan: Nothing.

bigflow/dataflow.py
iwko: Great! Thanks

test/test_dataflow.py
iwko: ok

bigflow/dataflow.py
iwko: Ok. Thx

bigflow/dataflow.py
iwko: 👍

bigflow/dataflow.py
iwko: sure or you meant not sure?

test/test_dataflow.py
anjensan: Yes. Sort of "do nothing" (because for most cases it is ok to have empty pipline).

bigflow/dataflow.py
anjensan: Yep, it mutates original 'pipeline_options['experiments']`list. Fixed.

bigflow/dataflow.py
anjensan: Do we need call super `__init__` ? Yes, we need this. But I forgot to add `**kwargs` fixed.

test/test_dataflow.py
anjensan: No. But it is passed by 'unittest.mock'

test/test_dataflow.py
iwko: So maybe rename it to do_nothing?

bigflow/dataflow.py
anjensan: Yes, there are few possible incompatibilities: 

1. `pipeline_options` might be `dict` or instance of `PipelineOptions`
in second case there is conversion `PipelineOptions->dict->PipelineOptions`, which potentially may cause problems.
I've added an [assertion](https://github.com/allegro/bigflow/pull/248/commits/b22e9e9935834aa1d2250bb2eaf8e9162b21095f) for this.

2. 'entry_point' might be 2-arg function when 'entry_point_arguments' is None.
Known difference - when entry_point has varargs (like `lambda pipeline, context, *arsg: ...`). In such case 3rd extra arg `{}` is not passed.

3. Some pipeline options are defaulted (aka 'job_name'), but only when they are not provided by user explicitly.
This should not break anything.

beczkowb (approved):

iwko (approved):

PR #250, opened 2021-04-27
Fix dataflow pipeline bool options parsing

anjensan:
Method `PipelineOptions.from_dictionary` just ignores all False values, but by default beam use public ips (so options 'no_use_private_ips=True' should be added to dict).  With this change 'use_private_ips=False' is handled properly.

### DIFF
--- a/bigflow/dataflow/__init__.py
+++ b/bigflow/dataflow/__init__.py
@@ -97,16 +97,15 @@ class BeamJob(Job):
 
         if isinstance(pipeline_options, PipelineOptions):
             logger.info("Convert PipelineOptions to dict")
+            orig = pipeline_options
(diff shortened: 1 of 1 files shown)


beczkowb (approved):

bigflow/dataflow/__init__.py
```diff
@@ -22,6 +22,9 @@
 
 logger = logging.getLogger(__file__)
 
+# PipelineOptions may dump unknown arguments to log which may cause passwords/secrets leakage.
+logging.getLogger("apache_beam.options.pipeline_options").setLevel(logging.ERROR)
```
beczkowb, line 26: +1

beczkowb (approved):

PR #251, opened 2021-04-27
Add experimental Konfig class

anjensan:
Experimental implementation, based on https://gist.github.com/anjensan/b74d2402a8dbd50caa0e72c5e4ce41c0
os.environ replacement and string expansion are explicit:
```python

class my_config(Konfig):
   password: str = fromenv("password")          # required, fail when is not available 
   pprefixed: str = expand("prefix_{password}") # expand placeholders
```


beczkowb:
Why `expand` is not the implicit default?

anjensan:
> Why `expand` is not the implicit default?

IMO it is less "magical" for a user.
It looks like a function call, which indicates "here is something happening, this is not just a string literal... your code will receive something different from what you see here"

Also technically call `expand` just defines a cached property, so it might be easier to get for someone who is experienced with python). And realisation is a bit simpler :)

beczkowb:
Is there a way to replace the custom Konfig class implementation with something like dataclass?

anjensan:
> Is there a way to replace the custom Konfig class implementation with something like dataclass? 

Yes, but with some limitations, see #253

anjensan:
Konfig name now should be passed as class parameter:

```python
class C(Konfig):
   x: int

class dev(C, name="dev"):
   x = 1

class prod(C, name="prod"):
   x = 2

resolve_konfig([dev, prod], 'prod')  
```

Or


```python
class C(Konfig):
   x: int

class _(C, name="dev"):
   x = 1

class _(C, name="prod"):
   x = 2

C.resolve('prod')  
```

bigflow/konfig.py
```diff
@@ -0,0 +1,280 @@
...
+    logger.info("Create instance of konfig %s", konfig_cls)
+    return konfig_cls(**extra)
+
+
+class secretstr(str):
+    def __repr__(self):
+        return f"<secretstr {'*' * len(self)}>"
+
+
+def fromenv(key: str, default=None, type: tp.Type = secretstr):
```
beczkowb, line 222: I think that the prefix can be another argument with the default value == 'bf_'. It gives more flexibility and it's explicit. If you want to have a default prefix in your project you can always create a partial.

beczkowb: And then you can delete the implicit mechanism

bigflow/konfig.py
```diff
@@ -0,0 +1,280 @@
...
+    return cached_property(get)
+
+
+_placeholder_re = re.compile(r"""
+    \{ (\w+) \}   # placeholder `{key}`
+    | \}\}        # literal `}}`
+    | \{\{        # literal `}}`
+""", re.VERBOSE)
+
+def _resolve_placeholders(value, resolve):
```
beczkowb, line 268: types hints

bigflow/konfig.py
```diff
@@ -0,0 +1,280 @@
...
+    """Creates instance of config.
+
+    Konfigs may be passed as list (in such case Class.name is used as a konfig name)
+    or as a plain dictionary {name -> konfig_class}.
+
+    Konfig name may be provided explicitly with `name` argument or passed via `bigflow` cli tool.
+    Default konfig config name may be passed as a fallback.
+    """
+
+    if not isinstance(konfigs, tp.Dict):
```
beczkowb, line 189: Im not sure if having 2 mechanisms for the same task is a good idea. I would pick the one with the argument in the class definition.

bigflow/konfig.py
```diff
@@ -0,0 +1,280 @@
...
+import re
+import lazy_object_proxy
+import typing as tp
+
+
+try:
+    from functools import cached_property
+except ImportError:
+    from backports.cached_property import cached_property
+
```
beczkowb, line 16: type hints please, so we can use mypy

bigflow/konfig.py
```diff
@@ -0,0 +1,280 @@
...
+    except KeyError:
+        raise ValueError(f"Unable to find konfig with name{name!r}, candidates {list(konfigs.keys())}")
+
+    logger.info("Create instance of konfig %s", konfig_cls)
+    return konfig_cls(**extra)
+
+
+class secretstr(str):
+    def __repr__(self):
+        return f"<secretstr {'*' * len(self)}>"
```
beczkowb, line 219: +1

beczkowb (changes requested):
needs work

bigflow/konfig.py
anjensan: I actually like this idea.
Maybe just completely remove this prefixing, user can always do it manually.
WDYT ?

bigflow/konfig.py
anjensan: I'd prefer to keep only 'dict' variant. It is a bit more flexible (you can always redefine konfigs), less magical,
and also clearly shows what environments do you have (no need to search through the code). WDYT?

bigflow/konfig.py
beczkowb: I think it's ok.

beczkowb: (removing it)

bigflow/konfig.py
beczkowb: Agree, keep the dict based mechanism

bigflow/konfig.py
anjensan: done

bigflow/konfig.py
anjensan: done

bigflow/konfig.py
anjensan: done

beczkowb (approved):

PR #253, opened 2021-05-10
Implement Konfig based on dataclasses

anjensan:
Reimplement 'Konfig' based on dataclasses.

Each class based on Konfig is turned into 'dataclass' (via custom metaclass).
All public non-descriptor values are converted to fields (there is not need to put type annotation). Also `cached_properties` (not generic properties) are handled in a special way, so they also treated as a fields (with full ability to redefine/inherit them).

Dataclasses give some abilities "for free":
 - immutability
 - method '.replace'
 - nice repr/str
 - better compat with other tools
 - override any prop via constructor

There are some drawbacks however:
 - it is not possible to mix fields with and without default value
    (error is thrown when field without defaul value follows a field with it)
 - magic is required to handle `dynamic_property` (see `Konfig.__getattribute__`)
 - there is `__post_init__`, but object is already frozen

### DIFF
--- a/bigflow/konfig.py
+++ b/bigflow/konfig.py
@@ -29,26 +29,30 @@ class KonfigMeta(abc.ABCMeta):
     @staticmethod
     def prepare_dict(dct: tp.Dict):
 
-        anns = dct.setdefault('__annotations__', {})
-        for k, v in list(dct.items()):
+        annotations = dct.setdefault('__annotations__', {})
+        for attr_name, attr in list(dct.items()):
 
-            if k.startswith("_"):
-                continue
-            elif isinstance(v, cached_property):
-                pass
-            elif hasattr(v, '__get__') or hasattr(v, '__set__') or hasattr(v, '__delete__'):
+            is_cached_property = isinstance(attr, cached_property)
+            is_descriptor = any(
+                hasattr(attr, f) for
+                f in ['__get__', '__set__', '__delete__']
+            )
+
+            if (
+                attr_name.startswith("_")
+                or (is_descriptor and not is_cached_property)
+            ):
                 continue
 
-            if k in anns:
-                pass
-            elif isinstance(v, cached_property):
-                anns[k] = tp.Any
-            else:
-                anns[k] = type(v)
+            if attr_name not in annotations:
+                if isinstance(attr, cached_property):
+                    annotations[attr_name] = tp.Any
+                else:
+                    annotations[attr_name] = type(v)
 
             # it is not allowed to use a mutable as a default value
-            if not isinstance(v, (tp.Hashable, cached_property)):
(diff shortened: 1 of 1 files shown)


bigflow/konfig.py
```diff
@@ -26,64 +27,58 @@ def current_env():
...
                 continue
-            elif hasattr(v, '__get__'):
-                # descriptor, pop out attribute from obj
-                obj.__dict__.pop(k, None)
-            else:
-                # normal value, put into obj
-                obj.__dict__[k] = v
-
-    def __new__(self, name, bases, dct):
+            elif isinstance(v, cached_property):
```
iwko, line 37: My cacheujemy jakoś properties?

bigflow/konfig.py
```diff
@@ -26,64 +27,58 @@ def current_env():
...
+                pass
+            elif hasattr(v, '__get__') or hasattr(v, '__set__') or hasattr(v, '__delete__'):
```
iwko, line 39: Co to za sprawdzenie?

anjensan: Descriptor test https://docs.python.org/3/howto/descriptor.html#descriptor-protocol

bigflow/konfig.py
anjensan: Helpers `expand` and `fromenv` are implemented based on `cached_property`.

bigflow/konfig.py
```diff
@@ -26,64 +27,58 @@ def current_env():
 class KonfigMeta(abc.ABCMeta):
 
     @staticmethod
-    def _copy_classvars_to_obj(klass, obj):
-        for k, v in klass.__dict__.items():
+    def prepare_dict(dct: tp.Dict):
+
+        anns = dct.setdefault('__annotations__', {})
+        for k, v in list(dct.items()):
```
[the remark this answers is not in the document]
anjensan: renamed to 'attr_name' and 'attr'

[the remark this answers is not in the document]
anjensan: refactored

anjensan:
It looks like dataclass-based implementation of 'Konfig' is bigger than naive implementation from #251 
Also it has one big disadvantage - dataclasses doesn't handle diamond inheritance well (which might be helpful in some cases). 

With original implementation:
```python
class config(Konfig):
    machine_num: int = 1
    machine_size: int = 1
    cores: int = dynamic(lambda s: s.machine_num * s.machine_size)

class big_machines(config):
    machine_size = 16

class test_conf(config):
    machine_num = 50

class test_big_machines(big_machines, test_conf):
    pass

test_big_machines()
# => test_big_machines(machine_num=1, machine_size=16, cores=16)
```

In contrast original implementation gives:
```python
# => <Config {'cores': 800, 'machine_num': 50, 'machine_size': 16}>
```

I'm closing this PR as it seems too tricky.

PR #266, opened 2021-06-20
removing bigflow.migrate

beczkowb:


bigflow/scaffold/scaffold.py
```diff
@@ -36,14 +36,4 @@ def start_project(
     )
 
     import bigflow.build.pip as bf_pip
-    bf_pip.pip_compile(project_path / "resources" / "requirements.txt")
-
-
-def migrate_project_from_10(
```
anjensan, line 42: What is motivation to remove this code?

test/test_migrate.py
```diff
@@ -1,44 +0,0 @@
...
-
-        p.read()  # ignore the rest of output
-        self.assertFalse(p.wait())
-
-        self.assertFileNotExists("project_setup.py")
-        self.assertFileExists("pyproject.toml")
-        self.assertFileExists("setup.py")
-
-
-class TestMigrateFromV11(_BaseTestCase):
```
anjensan, line 36: This test should not be removed.

anjensan (changes requested):

bigflow/scaffold/scaffold.py
beczkowb: 1. all projects inside allegro are already migrated
2. we are releasing version 1.3, I don't think we should keep migration tools forever
3. in connection to point 1, Im not sure if we have any users outside the allegro right now :(

PR #268, opened 2021-06-23
Update documentation

anjensan:


docs/configuration.md
```diff
@@ -326,4 +326,209 @@ deployment_config = DeploymentConfig(
...
-```
\ No newline at end of file
+```
+
+
+# Class-based configuration [experimental]
+
+Bigflow 1.3 introduces new class-based approach for keeping configurations.
+New API is optional and there is no need to migrate existing code from
+`biglfow.Configuration`. However it betters plays with typehinting, enables
```
beczkowb, line 336: However**,** it

beczkowb: However it betters plays with typehinting -> However, it allows you to use type hints, which enables...

docs/configuration.md
```diff
@@ -326,4 +326,209 @@ deployment_config = DeploymentConfig(
...
+class DevConf(CommonConf):
+    project = "dev-project"
+
+class ProdConf(CommonConf):
+    project = "prod-project"
+    num_workers = 20
+```
+
+Config which corresponds to the current environment may be created by function `resolve_konfig`.
+It takes dictionary in form of {"environment" -> "config class"} and returns instance of the config.
```
beczkowb, line 363: I think that you should mention that the configuration will be resolved based on the same environment variable mechanism, used in the "standard" bigflow.Configuration.

docs/configuration.md
```diff
@@ -326,4 +326,209 @@ deployment_config = DeploymentConfig(
        'docker_repository': 'your_repository',
        'gcp_project_id': '{project_id}',
        'dags_bucket': '{dags_bucket}'}})
-```
\ No newline at end of file
+```
+
+
+# Class-based configuration [experimental]
```
beczkowb, line 332: Use Grammarly to check the new docs.

bigflow/scaffold/templates/new-project/{{project_name}}/wordcount/pipeline.py.j2
```diff
@@ -14,26 +11,20 @@ workflow_config = Config(
...
-    options.view_as(WorkerOptions).max_num_workers = 2
-    options.view_as(WorkerOptions).autoscaling_algorithm = 'THROUGHPUT_BASED'
-    options.view_as(StandardOptions).runner = 'DataflowRunner'
-
-    options.view_as(SetupOptions).setup_file = str(materialize_setuppy().absolute())
-
-    logger.info(f"Run beam pipeline with options {str(options)}")
-    return options
\ No newline at end of file
+    return dict(
```
iwko, line 20: Maybe we should add here 'no_use_public_ips': True

docs/configuration.md
```diff
@@ -326,4 +326,209 @@ deployment_config = DeploymentConfig(
...
+Config which corresponds to the current environment may be created by function `resolve_konfig`.
+It takes dictionary in form of {"environment" -> "config class"} and returns instance of the config.
+
+```python
+config = bigflow.konfig.resolve_konfig(
+    {  # mapping of all configs
+        'dev': DevConf,
+        'prod': ProdConf,
+    },
+    default="dev",  # when no environment is porvided via CLI
```
iwko, line 371: porvided -> provided

docs/configuration.md
```diff
@@ -326,4 +326,209 @@ deployment_config = DeploymentConfig(
...
+class ProdConf(CommonConf):
+    project = "prod-project"
+    num_workers = 20
+```
+
+Config which corresponds to the current environment may be created by function `resolve_konfig`.
+It takes dictionary in form of {"environment" -> "config class"} and returns instance of the config.
+
+```python
+config = bigflow.konfig.resolve_konfig(
```
iwko, line 366: Cant we automatize this mapping somehow? It adds additional step to make and is error prone. Can't we require from every Konfig instance to contain `env` variable?

docs/configuration.md
```diff
@@ -326,4 +326,209 @@ deployment_config = DeploymentConfig(
...
+You need to be explicitly wrap interpolation pattern with function `bigflow.konfig.expand()`:
+
+```python
+from bigflow.konfig import expand
+
+class MyConfig(bigflow.konfig.Konfig):
+
+    docker_repository_project = "my-shared-docker-project-id"
+
+    # no 'expand()' function - string is *not* inerpolcatied
```
iwko, line 410: inerpolcatied -> interpolated

docs/configuration.md
```diff
@@ -326,4 +326,209 @@ deployment_config = DeploymentConfig(
...
+
+
+config = MyConfig()
+print(config.docker_repository_raw)  # => eu.gcr.io/{docker_repository_project}/my-analytics
+print(config.docker_repository)      # => eu.gcr.io/my-shared-docker-project-id/my-analytics
+```
+
+### Reading from environment variables
+
+Class-based configs system does not automatically read properties from OS environment.
```
iwko, line 424: Why it does not automatically read them?

docs/configuration.md
```diff
@@ -326,4 +326,209 @@ deployment_config = DeploymentConfig(
...
+
+Configuration properties may be declared as results of some expression:
+
+```python
+from bigflow.konfig import expand, dynamic
+
+class MyConfig(bigflow.konfig.Konfig):
+    num_machines = 10
+    num_cores = 4
+    machine_type = expand("n2-standard-{num_cores}")  # internally `epxand` also uses `dynamic`
```
iwko, line 506: epxand -> expand

iwko:
CI fails

docs/configuration.md
anjensan: Mostly because
> Explicit is better than implicit.

It might be hard to code reader to recognise that some property comes from environment.

docs/configuration.md
anjensan: Yes, we can, initial "versions" of `konfig.py` had ability to use list of konfig classes instead of dictionary.
Also there was method for auto-scan all subclasses (eg `config = CommonConf.resolve(default='dev')`.

But now it can be only dict (by design).  Mosty for making list of envs/configs explicit (you immediately see what environments you have in you project).

docs/configuration.md
beczkowb: done

docs/configuration.md
beczkowb: done

bigflow/scaffold/templates/new-project/{{project_name}}/wordcount/pipeline.py.j2
beczkowb: I think that in the public scaffold it's not a good idea, it's not that easy to set up networking to make it work

docs/configuration.md
beczkowb: done

docs/configuration.md
beczkowb: I agree with @anjensan. Having a single way of resolving configuration is better than having 2, and doing it through the dict mechanism is easier to understand and manipulate.

docs/configuration.md
beczkowb: done

docs/configuration.md
beczkowb: I opt for explicit solution

docs/configuration.md
beczkowb: done

iwko (approved):

PR #280, opened 2021-09-01
making docker image removal during build non-critical, adding more lo…

beczkowb:
…gging info


bigflow/commons.py
```diff
@@ -232,7 +232,9 @@ def decode_version_number_from_file_name(file_path: Path):
 
 
 def get_docker_image_id(tag):
+    logger.info("Getting docker image ID.")
     images = subprocess.getoutput(f"docker images -q {tag}")
```
anjensan, line 236: btw `bigflow.commons.run_process` put stdout/stderr to logger (with debug or info).

beczkowb: I have changed getoutput to run_process

PR #284, opened 2021-09-08
Hourly partitions

iwko:


e2e/test_dataset_manager.py
```diff
@@ -59,6 +60,92 @@ def tearDown(self):
...
+        self.internal_tables = [
+            'partitioned_hourly_fake_target_table',
+            'partitioned_daily_fake_target_table'
+        ]
+
+    def tearDown(self):
+        self.dataset_manager.remove_dataset()
+
+    def test_should_collect_records_from_hourly_partitioned_table(self):
+        self.test_dataset_id, self.dataset_manager = create_dataset_manager(
```
beczkowb, line 78: The partitioning type shouldn't be set on the dataset level, you can have multiple tables in a dataset with multiple types of partitioning (daily, hourly, clustered). The partitioning type should be configurable on the level of an operation – write_truncate(partitioning_type=...), write_append(...), etc.

PR #296, opened 2021-11-03
Updade dag-generator - use Airflow 2.x paths, refactor

anjensan:
Update generated DAG files - try Airflow 2.x paths and fallback to 1.x on ImportError.
Also code is slightly refactored - f-strings are used instead of `str.format`.
Added comments to generated files - worflow id / image version / etc.


beczkowb (approved):

anjensan:
tested - DAGs work on composer1 and composer2
however composer2 uses different auth model - some extra steps might be required to configure a cluster https://cloud.google.com/composer/docs/composer-2/access-control#composer-sa

PR #297, opened 2021-11-05
Bump max beam version to 2.33, fix deps

anjensan:


requirements/dataproc_extras.txt
```diff
@@ -1,2 +1,7 @@
-google-cloud-dataproc>=2,<3
-pyspark>=3.0,<4
\ No newline at end of file
+google-cloud-dataproc>=3,<4
```
beczkowb, line 1: why did you bump dataproc?

PR #298, opened 2021-11-11
Propagate env to Dataflow workers

anjensan:
Propagate current env to beam worker via PipelineOptions:
 - auto-add 'bigflow_env' to pipeline options
 - inspect not only os.environ['bf_env'], but also check pipeline options (when they are available)
 - support both DirectRunner and DataflowRunner
 - no unit tests to 'get_pipeline_options' as this function should be testes e2e on real Dataflow (TBD)


bigflow/configuration.py
```diff
@@ -32,6 +27,12 @@ def __init__(self,
 
         self.add_configuration(name, properties, is_default)
 
+    def _current_env(self):
```
beczkowb, line 30: +1

PR #299, opened 2021-11-26
Revert "Bump max beam version to 2.33, fix deps (#297)"

iwko:
This reverts commit fc12ff919fad41d4213dfc3c78efa53b4b489f1f.


anjensan:
Probably better to revert only 2030823462fb1a505a598896c38f36932c190e1d ?

anjensan (changes requested):
I suppose it is better to revert only 2030823

Issue #242, opened 2021-04-15
Switch to 'pytest' as a runner for unit tests


anjensan:
Pytest is much more user-friendly.

1. Better runner - colors, stacktraces, code snipets, hides "garbage" stdout/err/logger output.
2. Concise tests - just `def test(): assert ...` instead of junit-approach
3. Tons! of [useful plugins ](https://docs.pytest.org/en/latest/reference/plugin_list.html); ability to create bigflow-specific.

PR #301, opened 2021-12-04
Add pytest support #242

anjensan:


beczkowb (changes requested):
* Missing documentation – how to enable pytest
* Missing comment – why do we need the _ConsoleStreamLogHandler class

beczkowb (approved):

PR #304, opened 2021-12-14
dataset & table labels in DatasetConfig

iwko:


bigflow/bigquery/dataset_manager.py
```diff
@@ -309,9 +311,20 @@ def create_table(self, create_query):
         job = self.bigquery_client.query(
             create_query,
             job_config=job_config)
-        return job.result()
+        job_result = job.result()
```
beczkowb, line 314: I'm wondering if that is the best place to apply labels because not all tables must be created using `create_table`. It's also confusing because you might think that if you declare labels in the dataset configuration, labels will be applied "no matter what". I suggest alternative solutions:

- creating table labels in the `create_dataset_manager` function
- adding an explicit, optional argument to the `create_table` method
- creating new method on dataset manager, called `upsert_table_labels`

also, @anjensan what do you think?

iwko: I was thinking about all solutions you've mentioned but I implemented labels in this way because:
> creating table labels in the create_dataset_manager function
If the table does not exist yet then we can't add a label.

>adding an explicit, optional argument to the create_table method
User must provide labels for every call of create_table which is uncomfortable 

>creating new method on dataset manager, called upsert_table_labels
It adds another separate step and users need to remember about adding it.

Maybe adding a prefix to param names will be enough?

` tables_labels, dataset_labels` -> ` create_table_labels, create_dataset_labels`

beczkowb: "If the table does not exist yet then we can't add a label." -> is it a problem?

iwko: > "If the table does not exist yet then we can't add a label." -> is it a problem?

Yes, because then we will need to define labels during creating a table so we will be touching labels in two places

beczkowb: can't we just do something like this in `create_dataset_manager`?:

```python
if table_exists:
    upsert_labels()
```

beczkowb: (and don't do anything in `create_table`)

iwko: But if a table does not exist then we have to pass labels in `create_table` method. So instead of passing labels only to `create_dataset_manager` we have to pass in two places.

beczkowb: I think we can just have the logic in the `create_dataset_manager` method and assume that if that's the first run, before `create_table`, there will be no labels until the second run

iwko: Ok, will change to that.

iwko: See now

bigflow/bigquery/dataset_manager.py
```diff
@@ -408,37 +408,63 @@ def _convert_to_humanbytes(size_in_bytes):
                 'cost': cost}
 
 
-def create_dataset(dataset_name, bigquery_client, location=DEFAULT_LOCATION):
+def create_dataset(dataset_name: str, bigquery_client, location: str = DEFAULT_LOCATION, dataset_new_labels: Dict = None):
```
beczkowb, line 411: Dict[str, str]?

bigflow/bigquery/dataset_manager.py
```diff
@@ -408,37 +408,63 @@ def _convert_to_humanbytes(size_in_bytes):
...
-    return bigquery_client.create_dataset(dataset, exists_ok=True)
+    bigquery_dataset = bigquery_client.create_dataset(dataset, exists_ok=True)
+    if dataset_new_labels:
+        logger.info(f'ADDING LABELS FOR DATASET: ${dataset_name}')
+        bigquery_dataset.labels = _prepare_labels(bigquery_dataset.labels, dataset_new_labels)
+        return bigquery_client.update_dataset(bigquery_dataset, ["labels"])
+    return bigquery_dataset
+
+
+def _prepare_labels(labels: Dict, new_labels: Dict):
```
beczkowb, line 425: more specific typing please

beczkowb: also, add the return type

docs/technologies.md
```diff
@@ -466,25 +467,29 @@ workflow = Workflow(
         schedule_interval='@once')
 ```
 
-#### Label table
+#### Labels
 
-The `add_label` function allows your workflow to create/overrides a label for a BigQuery table.
+The `table_labels` and `dataset_labels` parameters allows your workflow to create/overrides a label for a BigQuery table and dataset.
```
beczkowb, line 472: pass it through Grammarly, I'm not sure if it's grammatically correct

e2e/test_dataset_manager.py
```diff
@@ -185,8 +226,82 @@ def test_should_create_table(self):
...
+        CREATE TABLE labeled_table (
+            id STRING
+        )''')
+
+        # then
+        self.assertEqual(
+            self.dataset_manager.client.get_table(self.test_dataset_id + '.' + 'labeled_table').labels, {}
+        )
+
+    def test_should_upsert_labels_when_table_exists(self):
```
beczkowb, line 266: test_should_upsert_labels_when_table_exists -> test_should_upsert_labels?

e2e/test_dataset_manager.py
```diff
@@ -185,8 +226,82 @@ def test_should_create_table(self):
...
+            dataset_labels={'test_label_key': 'test_label_value', 'another_test_label_key': 'another_test_label_value'},
+            tables_labels={'labeled_table': {'labeled_table_key': 'labeled_table_value', 'another_labeled_table_key': 'another_labeled_table_value'}})
+
+        # then
+        self.assertEqual(
+            dataset_manager.client.get_table(test_dataset_id + '.' + 'labeled_table').labels,
+            {'labeled_table_key': 'labeled_table_value', 'another_labeled_table_key': 'another_labeled_table_value'}
+        )
+
+    def test_should_not_add_labels(self):
```
beczkowb, line 253: I don't get the scenario this test is checking, can you explain?

e2e/test_dataset_manager.py
```diff
@@ -59,7 +61,46 @@ def tearDown(self):
...
+            config.PROJECT_ID,
+            self.TEST_PARTITION,
+            dataset_name="without_labels"+self.dataset_uuid,
+            internal_tables=self.internal_tables,
+            external_tables=self.external_tables)
+
+        # then
+        self.assertEqual(self.dataset_manager.client.get_dataset(self.test_dataset_id).labels, {})
+
+    def test_should_upsert_labels_when_dataset_exists(self):
```
beczkowb, line 82: test_should_upsert_labels?

e2e/test_dataset_manager.py
```diff
@@ -59,7 +61,46 @@ def tearDown(self):
...
 
 
-class PartitionedDatasetManagerPropertiesTestCase(DatasetManagerTestCase):
+class DatasetManagerTestCase(DatasetManagerBaseTestCase):
+    def test_should_have_labels(self):
+        # expect
+        self.assertEqual(self.dataset_manager.client.get_dataset(self.test_dataset_id).labels,
+                         {'test_label_key': 'test_label_value', 'another_test_label_key': 'another_test_label_value'})
+
+    def test_should_not_have_labels(self):
```
beczkowb, line 70: test_should_not_add_labels?

e2e/test_dataset_manager.py
```diff
@@ -59,7 +61,46 @@ def tearDown(self):
         self.dataset_manager.remove_dataset()
 
 
-class PartitionedDatasetManagerPropertiesTestCase(DatasetManagerTestCase):
+class DatasetManagerTestCase(DatasetManagerBaseTestCase):
+    def test_should_have_labels(self):
```
beczkowb, line 65: test_should_add_labels?

beczkowb, line 65: I don't understand this test, can you explain?

e2e/test_dataset_manager.py
iwko: It checks that on the first call there are no labels. It is more like document-test then real testing. I can remove this test. WDYT?

e2e/test_dataset_manager.py
iwko: It tests if after creating dataset this dataset has labels

e2e/test_dataset_manager.py
iwko: Done

e2e/test_dataset_manager.py
iwko: Done

e2e/test_dataset_manager.py
iwko: Done

e2e/test_dataset_manager.py
iwko: Done

docs/technologies.md
iwko: Done, but for better correction Grammarly Premium is needed :D

bigflow/bigquery/dataset_manager.py
iwko: Done

bigflow/bigquery/dataset_manager.py
iwko: Done

e2e/test_dataset_manager.py
beczkowb: remove it then

e2e/test_dataset_manager.py
beczkowb: then call it `test_should_add_labels`

beczkowb (approved):

PR #306, opened 2022-01-14
FST-250 bf build arguments validation

artnowo-alle:
* moved `_valid_datetime` to `bigflow.commons` as `valid_datetime`
* shuffled activities order in `bf build` (was: package -> image -> dags, is: dags -> package -> image)
* moved workflow loading in `operate.build_dags` before other activities
* made `build`/`build-dags` commands fail when no workflows found for `ALL`/none workflow ID

### DIFF
--- a/bigflow/build/operate.py
+++ b/bigflow/build/operate.py
@@ -114,17 +114,11 @@ def build_dags(
     start_time: str,
     workflow_id: typing.Optional[str] = None,
 ):
-    logger.info("Building airflow DAGs...")
-    clear_dags_leftovers(project_spec)
+    # TODO: Move common functions from bigflow.cli to bigflow.commons (or other shared module)
+    from bigflow.cli import walk_workflows
 
-    image_version = bf_commons.build_docker_image_tag(project_spec.docker_repository, project_spec.version)
-    create_image_version_file(str(project_spec.project_dir), image_version)
-
-    # TODO: Move common frunctions from bigflow.cli to bigflow.commons (or other shared module)
-    from bigflow.cli import _valid_datetime, walk_workflows
-    _valid_datetime(start_time)
-
-    cnt = 0
+    logger.debug('Loading workflow(s)...')
+    workflows = []
     for root_package in project_spec.packages:
         if "." in root_package:
             # leaf package
@@ -133,18 +127,32 @@ def build_dags(
         for workflow in walk_workflows(project_spec.project_dir / root_package):
             if workflow_id is not None and workflow_id != workflow.workflow_id:
                 continue
-            logger.info("Generating DAG file for %s", workflow.workflow_id)
-            cnt += 1
-            bigflow.dagbuilder.generate_dag_file(
-                str(project_spec.project_dir),
-                image_version,
-                workflow,
-                start_time,
-                project_spec.version,
-                root_package,
-            )
+            workflows.append((workflow, root_package))
 
-    logger.info("Geneated %d DAG files", cnt)
+    if not workflows:
+        if not workflow_id:
+            raise Exception('No workflow found')
+        else:
+            raise Exception("Workflow '{}' not found".format(workflow_id))
+
+    logger.info("Building airflow DAGs...")
+    clear_dags_leftovers(project_spec)
+
+    image_version = bf_commons.build_docker_image_tag(project_spec.docker_repository, project_spec.version)
+    create_image_version_file(str(project_spec.project_dir), image_version)
+
+    for (workflow, package) in workflows:
+        logger.info("Generating DAG file for %s", workflow.workflow_id)
+        bigflow.dagbuilder.generate_dag_file(
+            str(project_spec.project_dir),
+            image_version,
+            workflow,
+            start_time,
+            project_spec.version,
+            package,
+        )
+
+    logger.info("Generated %d DAG files", len(workflows))
 
 
 def _rmtree(p: Path):
@@ -194,7 +202,7 @@ def build_project(
     workflow_id: typing.Optional[str] = None,
 ):
     logger.info("Build the project")
+    build_dags(project_spec, start_time, workflow_id=workflow_id)
     build_package(project_spec)
     build_image(project_spec)
-    build_dags(project_spec, start_time, workflow_id=workflow_id)
     logger.info("Project was built")
--- a/bigflow/cli.py
+++ b/bigflow/cli.py
@@ -303,21 +303,6 @@ def _create_build_package_parser(subparsers):
     subparsers.add_parser('build-package', description='Builds .whl package from local sources.')
 
 
-def _valid_datetime(dt):
-    if dt == 'NOW':
-        return
-
-    try:
-        datetime.strptime(dt, "%Y-%m-%d %H:%M:%S")
-        return dt
-    except ValueError:
-        try:
-            datetime.strptime(dt, "%Y-%m-%d")
-            return dt
-        except ValueError:
-            raise ValueError("Not a valid date: '{0}'.".format(dt))
-
-
 def _add_build_dags_parser_arguments(parser):
     parser.add_argument('-w', '--workflow',
                         type=str,
@@ -330,7 +315,7 @@ def _add_build_dags_parser_arguments(parser):
                              'For workflows triggered hourly -- datetime in format: Y-m-d H:M:S, for example 2020-01-01 00:00:00. '
                              'For workflows triggered daily -- date in format: Y-m-d, for example 2020-01-01. '
                              'If empty or set as NOW, current hour is used.',
-                        type=_valid_datetime)
+                        type=bf_commons.valid_datetime)
 
 
 def _create_build_dags_parser(subparsers):
--- a/bigflow/commons.py
+++ b/bigflow/commons.py
@@ -320,3 +320,24 @@ def as_timedelta(v: None | str | Number | timedelta) -> timedelta | None:
         return None
     else:
         return timedelta(seconds=float(v))
+
+
+def valid_datetime(dt: str) -> str:
+    """
+    Validates provided datetime string. Raises ValueError for strings that are none of:
+    * 'NOW'
+    * valid '%Y-%m-%d %H:%M:%S'
+    * valid '%Y-%m-%d'
+    """
+    if dt == 'NOW':
+        return dt
+
+    try:
+        datetime.strptime(dt, "%Y-%m-%d %H:%M:%S")
+    except ValueError:
+        try:
+            datetime.strptime(dt, "%Y-%m-%d")
+        except ValueError:
+            raise ValueError("Not a valid date: '{0}'.".format(dt))
+
+    return dt
--- a/test/bf-projects/bf_simple_pytest/bf_simple_pytest/workflow.py
+++ b/test/bf-projects/bf_simple_pytest/bf_simple_pytest/workflow.py
@@ -1,0 +1,6 @@
+import bigflow
+
+the_workflow = bigflow.Workflow(
+    workflow_id='the_workflow',
+    definition=[]
+)
--- a/test/test_commons.py
+++ b/test/test_commons.py
@@ -43,6 +43,45 @@ class CliTestCase(TestCase):
             # when
             decode_version_number_from_file_name(Path('/Users/image-0.1122123.0.tar'))
 
+    def test_valid_datetime_should_pass_for_NOW(self):
+        # when
+        valid_datetime('NOW')
+
+        # then
+        # no error is raised
+
+    def test_valid_datetime_should_pass_for_valid_YmdHMS(self):
+        # when
+        valid_datetime('2022-01-01 12:34:56')
+
+        # then
+        # no error is raised
+
+    def test_valid_datetime_should_raise_error_for_invalid_YmdHMS(self):
+        # then
+        with self.assertRaises(ValueError):
+            # when
+            valid_datetime('2022-01-01 34:56:78')
+
+    def test_valid_datetime_should_pass_for_valid_Ymd(self):
+        # when
+        valid_datetime('2022-01-01')
+
+        # then
+        # no error is raised
+
+    def test_valid_datetime_should_raise_error_for_invalid_Ymd(self):
+        # then
+        with self.assertRaises(ValueError):
+            # when
+            valid_datetime('2022-23-45')
+
+    def test_valid_datetime_should_raise_error_for_other_string(self):
+        # then
+        with self.assertRaises(ValueError):
+            # when
+            valid_datetime('foo bar baz')
+
     def _touch_file(self, file_name: str, content: str = ''):
         workdir = Path(os.getcwd())
         f = workdir / file_name


bigflow/cli.py
```diff
@@ -117,6 +116,28 @@ def walk_workflows(root_package: Path) -> Iterator[bf.Workflow]:
             yield workflow
 
 
+def walk_workflows_in_packages(
+        project_dir: Path,
+        packages: List[str],
+        predicate: Callable[[bf.Workflow], bool] = lambda w: True,
```
iwko, line 122: Is there a test for predicate == False?

bigflow/cli.py
```diff
@@ -117,6 +116,28 @@ def walk_workflows(root_package: Path) -> Iterator[bf.Workflow]:
...
+        transformer: Callable[[bf.Workflow], Any] = lambda w: w
```
iwko, line 123: I do not see use case for this parameter

test/cli/test_cli.py
```diff
@@ -114,6 +117,45 @@ def test_should_walk_through_all_workflows_inside_package_tree(self):
...
+
+        # then
+        self.assertEmpty(res)
+
+    def test_walk_workflows_in_packages_should_walk_through_all_workflows_without_transformer(self):
+        # when
+        res = walk_workflows_in_packages(TEST_PROJECT_DIR, ['test_module'])
+
+        # then
+        self.assertEqual([('ID_1', '@once'), ('ID_2', '@daily'), ('ID_3', '@daily'), ('ID_4', '@daily'), ('ID_5', '@daily')],
```
iwko, line 132: Why do we use map here and chekcs only `workflow_id` and `schedule_interval`

test/mixins.py
```diff
@@ -245,6 +245,14 @@ def bigflow_spawn(self, cmd, **kwargs):
         return self.subprocess_spawn(cmd, **kwargs)
 
 
+class IteratorAssertionsMixin(Mixin):
+    def assertEmpty(self, iterator: Iterator):
```
anjensan, line 249: It is better to use `Iterable` (well, it is only a typehint, so doesn't matter too much).
BTW any iterator should be an `Iterable` in Python.

Also add parameter `msg` to the function with default value `"Iterable should be empty"`, so user may
redefine that message.

test/mixins.py
```diff
@@ -245,6 +245,14 @@ def bigflow_spawn(self, cmd, **kwargs):
...
+        self.assertFalse(list(iterator), msg="Iterator should be empty")
```
anjensan, line 250: IMO those functions are not super needed (they save us from typing a few chars), and may be removed.

test/cli/test_cli.py
```diff
@@ -114,6 +117,45 @@ def test_should_walk_through_all_workflows_inside_package_tree(self):
         self.assertEqual('ID_2', res[1].workflow_id)
         self.assertNotEqual('@once', res[1].schedule_interval)
 
+    def test_walk_workflows_in_packages_should_not_walk_when_no_packages_given(self):
```
anjensan, line 120: Empty packages list should be considered as an error.

test/cli/test_cli.py
```diff
@@ -114,6 +117,45 @@ def test_should_walk_through_all_workflows_inside_package_tree(self):
...
+
+    def test_walk_workflows_in_packages_should_walk_through_all_workflows_without_transformer(self):
+        # when
+        res = walk_workflows_in_packages(TEST_PROJECT_DIR, ['test_module'])
+
+        # then
+        self.assertEqual([('ID_1', '@once'), ('ID_2', '@daily'), ('ID_3', '@daily'), ('ID_4', '@daily'), ('ID_5', '@daily')],
+            sorted(map(lambda w: (w.workflow_id, w.schedule_interval), res)))
+
+    def test_walk_workflows_in_packages_should_walk_through_all_workflows_inside_single_package(self):
```
anjensan, line 135: I'm pretty sure that "package walking" is already covered in our unit tests (like `test_should_walk_through_all_module_paths_inside_package_tree`). So maybe update existing tests to use new function (as it heavily depends on existing functions). 
Or at least reuse exiting "tests modules" (aka `Unused1.py` etc)

test/cli/test_cli.py
anjensan, line 132:
```python
        self.assertSetEqual(
            {('ID_1', '@once'), ('ID_2', '@daily'), ('ID_3', '@daily'), ('ID_4', '@daily'), ('ID_5', '@daily')},
            {(w.workflow_id, w.schedule_interval) for w in res},
        )
```

test/cli/test_cli.py
```diff
@@ -114,6 +117,45 @@ def test_should_walk_through_all_workflows_inside_package_tree(self):
...
+        res = walk_workflows_in_packages(TEST_PROJECT_DIR, ['test_module', 'other_test_module'],
+            transformer=lambda w: w.workflow_id)
+
+        # then
+        self.assertEqual(['ID_1', 'ID_2', 'ID_3', 'ID_4', 'ID_5', 'ID_9'], sorted(res))
+
+    def test_walk_workflows_in_packages_should_return_workflows_matching_predicate_only(self):
+        # when
+        res = walk_workflows_in_packages(TEST_PROJECT_DIR, ['test_module', 'other_test_module'],
+            lambda w: w.workflow_id.endswith('9') or w.workflow_id.endswith('2'),
```
anjensan, line 153: nit: `w.workflow_id[-1] in "92"`

test/cli/test_cli.py
```diff
@@ -135,6 +177,51 @@ def test_should_not_find_non_existing_workflow(self):
...
+        # then
+        with self.assertRaises(Exception):
+            # when
+            verify_workflow('ID_999', ['broken_workflows'], project_dir=TEST_PROJECT_DIR)
+
+    def test_verify_workflow_should_raise_error_for_workflow_with_syntax_error(self):
+        # then
+        with self.assertRaises(Exception):
+            # when
+            verify_workflow('ID_998', ['broken_workflows'], project_dir=TEST_PROJECT_DIR)
```
anjensan, line 217: Now we fail the whole build when there are any import issue (this include wrong import, global-thrown exception etc).  So basically all tests with `broken_workflows` do the same thing - fail at the first wrong worfklow.

In any case I don't see a big reasons to add those tests (as this is already covered).

anjensan (changes requested):

test/cli/test_cli.py
artnowo-alle: I'm mapping in order to avoid comparing whole objects. I followed existing `test_should_walk_through_all_workflows_inside_package_tree` test that asserts on ID and schedule interval.

test/mixins.py
artnowo-alle: `assertEmpty()` is way more readable then `assertFalse(list(...))`, so I'll insist on leaving them.

test/cli/test_cli.py
artnowo-alle: Nice!

test/cli/test_cli.py
artnowo-alle: I wrote them for my own purpose, actually, to confirm that it fails uniformly for different kinds of problems.

test/mixins.py
anjensan: This is `nit` for me, so you really could leave those functions if you think they are useful.

However I'm not fully agree that it is more readable, at least not in a Python.
"Readable" means "most of readers will understand this quickly and without extra brain-work".

There is an existing question for similar problem on SO https://stackoverflow.com/questions/33216488/how-to-assert-that-an-iterable-is-not-empty-on-unittest (assert list is empty, not an iterator).

**115** users vote for just using `self.assertTrue(d)` (and rely on well known python [truth/falsy](https://docs.python.org/3/library/stdtypes.html#truth-value-testing) concept).
**8** users vote for using/adding `self.assertEmpty`.

bigflow/cli.py
```diff
@@ -127,6 +148,41 @@ def find_workflow(root_package: Path, workflow_id: str) -> bf.Workflow:
             return workflow
     raise ValueError('Workflow with id {} not found in package {}'.format(workflow_id, root_package))
 
+
+def verify_workflow(workflow_id: str, packages: List[str] = None, project_dir: Path = None) -> None:
```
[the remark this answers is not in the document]
artnowo-alle: `find_workflow` looks up for a workflow with particular workflow_id. `verify_workflow` supports `ALL`/`None` workflow id (i.e. when `--workflow` is not set)

artnowo-alle:
Pushed WIP commit, most conversations resolved

test/cli/test_cli.py
artnowo-alle: Reused wherever possible, added just another package with correct modules and a special package with broken modules.

bigflow/build/operate.py
```diff
@@ -114,15 +115,18 @@ def build_dags(
     start_time: str,
     workflow_id: typing.Optional[str] = None,
 ):
+    logger.debug('Verifying workflow(s)...')
+    bf_workflows.verify_workflow(workflow_id)
```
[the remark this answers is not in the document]
artnowo-alle: Let's leave workflow-related refactorings for #244

bigflow/commons.py
```diff
@@ -320,3 +320,28 @@ def as_timedelta(v: None | str | Number | timedelta) -> timedelta | None:
...
+
+
+def valid_datetime(dt: str) -> str:
+    """
+    Validates provided datetime string. Raises ValueError for strings that are none of:
+    * 'NOW'
+    * valid '%Y-%m-%d %H:%M:%S'
+    * valid '%Y-%m-%d'
+    """
+    if dt == 'NOW':
```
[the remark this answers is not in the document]
artnowo-alle: It's used by `operate.build_dags()` with a TODO saying it should be moved out of it, so I won't move it back. For the same reason, I'd rather not move the `NOW` check nor split it in two parts.

bigflow/commons.py
```diff
@@ -320,3 +320,28 @@ def as_timedelta(v: None | str | Number | timedelta) -> timedelta | None:
...
+        try:
+            datetime.strptime(dt, "%Y-%m-%d")
+        except ValueError:
+            raise ValueError("Not a valid date: '{0}'.".format(dt))
+
+    return dt
+
+
+def is_workflow_id_specified(workflow_id: str) -> bool:
+    return workflow_id and workflow_id != 'ALL'
```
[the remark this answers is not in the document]
artnowo-alle: Not really a part of CLI IMO. It verifies plain string and is extracted from a function the verifies `args.workflow` that is a part of CLI indeed.

test/test_workflows.py
```diff
@@ -0,0 +1,116 @@
...
+        get_project_spec.return_value.project_dir = TESTS_DIR
+
+        # then
+        with self.assertRaises(Exception):
+            # when
+            verify_workflow('I DONT EXIST')
+
+    @mock.patch('bigflow.build.spec.get_project_spec')
+    def test_verify_workflow_should_raise_error_for_broken_workflow_without_specified_id(self, get_project_spec):
+        # given
```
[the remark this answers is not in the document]
iwko: I agree with @anjensan

bigflow/commons.py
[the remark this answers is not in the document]
iwko: I think `commons.py` is a good place for `valid_datetime` method. It is used by, `operate.py` and `cli.py`.

bigflow/build/operate.py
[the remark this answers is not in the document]
iwko: I like the idea of moving it outside this PR.

bigflow/cli.py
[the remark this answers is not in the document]
iwko: Maybe we could add support for ALL/None workflow id (i.e. when --workflow is not set) to find_workflow?

bigflow/workflows.py
```diff
@@ -0,0 +1,66 @@
+"""
```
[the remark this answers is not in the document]
iwko: We should merge those two

bigflow/commons.py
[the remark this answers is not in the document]
iwko: Would leave it here. Used bye two modules `cli.py` and `workflows.py`

bigflow/workflows.py
```diff
@@ -0,0 +1,66 @@
...
+    except Exception as e:
+        if bf_commons.is_workflow_id_specified(workflow_id):
+            raise Exception(f'Error while loading workflow {workflow_id!r}') from e
+        else:
+            raise Exception('Error while loading workflows') from e
+
+    if bf_commons.is_workflow_id_specified(workflow_id) and not workflows:
+        raise Exception(f'Workflow {workflow_id!r} not found')
+
+    for w in workflows:
```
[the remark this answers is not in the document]
iwko: I think we could merge this method with `find_workflow`

bigflow/workflow.py
```diff
@@ -417,3 +420,55 @@ def _parse_runtime_str(runtime: str) -> dt.datetime:
         except ValueError:
             pass
     raise ValueError("Unable to parse 'runtime' %r" % runtime)
+
+
+def load_workflows_in_packages(
+        project_dir: Path,
+        packages: List[str]
+) -> Iterator[Any]:
```
beczkowb, line 428: If you load workflows then it should be `Iterator[Workflow]`, am I right?

And I think that we don't need that docstring, it's basically describing the function signature.

bigflow/workflow.py
```diff
@@ -417,3 +420,55 @@ def _parse_runtime_str(runtime: str) -> dt.datetime:
...
+    return (
+        w
+        for p in packages if '.' not in p
+        for w in walk_workflows(project_dir / p)
+    )
+
+
+def verify_workflow(workflow_id: str) -> None:
+    """
+    Verifies that workflow with id workflow_id is present and can be correctly loaded, raises exception otherwise.
```
beczkowb, line 446: please, pass all kinds of documentation through grammarly, it's an easy way of fixing some basic errors, like missing a/an/the

bigflow/workflow.py
```diff
@@ -417,3 +420,55 @@ def _parse_runtime_str(runtime: str) -> dt.datetime:
...
+    # TODO move walk_workflows from cli to workflows
+    from bigflow.cli import walk_workflows
+    return (
+        w
+        for p in packages if '.' not in p
+        for w in walk_workflows(project_dir / p)
+    )
+
+
+def verify_workflow(workflow_id: str) -> None:
```
beczkowb, line 444: If I understand correctly, if I want to validate all workflows, then I need to pass some random string.

Maybe just make the workflow_id optional, and if it's None, then set a random string inside the method, it's easier to understand.

anjensan: AFAIK now it should be `"ALL"` or `""` (empty string)  to validate all workflows. `None` also works (despite the typehint), because `bigflow.commons.is_workflow_id_specified` also implicitly check for `None`.

BTW magic `"ALL"` is replaced by `None` in `cli.py` https://github.com/allegro/bigflow/blob/233fe90c039cec72f44601cc7a6f578926b884e6/bigflow/cli.py#L570
so `verify_workflow` is never called with an empty string nor the `"ALL"`

bigflow/workflow.py
artnowo-alle: Yeah, this is a leftover from previous version that was able to transform workflows to anything.

bigflow/workflow.py
artnowo-alle: Good catch, this was not intended ofc. Will fix during the current refactoring

anjensan: BTW. the same with "--start-time", it is replaced by `datetime.now().strftime("%Y-%m-%d %H:00:00")` in cli.py
This is what i meant when was saying "NOW and ALL are parts of cli". Sorry, it was not written clearly.

bigflow/workflows.py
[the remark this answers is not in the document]
artnowo-alle: Finally I decided not to do so due to contradictory needs of building and execution.

test/cli/test_cli.py
artnowo-alle: Finally decided to drop these tests.

PR #308, opened 2022-01-26
Add caching/exporting flags to 'bf build' and 'bf build-image'

anjensan:


bigflow/cli.py
```diff
@@ -318,15 +320,41 @@ def _add_build_dags_parser_arguments(parser):
...
 
 
+def _add_build_image_parser_arguments(parser: argparse.ArgumentParser):
+    parser.add_argument(
+        '--export-image-tar', dest='export_image_tar', action='store_true',
+        help="Export built docker image into .tar file",
+    )
+    parser.add_argument(
+        '--no-export-image-tar', dest='export_image_tar', action='store_false',
+        help="Don't epxort built docker image into .tar file (keep image in local docker registry)",
```
iwko, line 330: epxort -> export

bigflow/cli.py
```diff
@@ -318,15 +320,41 @@ def _add_build_dags_parser_arguments(parser):
...
 
 def _create_build_image_parser(subparsers):
-    subparsers.add_parser('build-image',
-                          description='Builds a docker image from local files.')
+    parser = subparsers.add_parser(
+        'build-image',
+        description='Builds a docker image from local files.',
+    )
+    _add_build_image_parser_arguments(parser)
+    _add_parsers_common_arguments(parser)
```
iwko, line 357: Do we have to add a common parser?

bigflow/cli.py
```diff
@@ -365,7 +393,7 @@ def _add_parsers_common_arguments(parser):
                              ' individual workflows as well as to deployment_config.py.')
 
 
-def _add_deploy_parsers_common_arguments(parser):
+def _add_dockerrepo_auth_parsers_arguments(parser):
```
iwko, line 396: _add_auth_parsers_arguments ?

bigflow/cli.py
```diff
@@ -485,20 +516,23 @@ def _resolve_vault_endpoint(args):
...
-        return config.resolve_property(property_name, args.config)
+def _resolve_property(args, property_name, ignore_value_error=False):
+    try:
+        cli_atr = getattr(args, property_name)
+        if cli_atr:
+            return cli_atr
+        else:
+            config = import_deployment_config(_resolve_deployment_config_path(args), property_name)
+            return config.resolve_property(property_name, args.config)
+    except ValueError:
```
iwko, line 527: Why do you want to allow ignoring missing properties?

bigflow/build/spec.py
```diff
@@ -84,6 +85,7 @@ def parse_project_spec(
     project_requirements_file="resources/requirements.txt",
     resources_dir="resources",
     test_framework='unittest',
+    export_image_tar=True,
```
iwko, line 88: It wouldn't be better to use `False` as a default?

bigflow/build/operate.py
```diff
@@ -88,18 +138,44 @@ def build_image(
...
+    _build_docker_image(project_spec, tag, cache_params)
 
-    try:
-        _export_docker_image_to_file(tag, image_dir, project_spec.version)
-        dconf_file = Path(project_spec.deployment_config_file)
-        shutil.copyfile(dconf_file, image_dir / dconf_file.name)
-    finally:
-        logger.info("Trying to remove the docker image. Tag: %s, image ID: %s", tag, bf_commons.get_docker_image_id(tag))
+    if export_image_tar:
         try:
```
TSienki, line 144: Extracting this if (and else) content to separate functions wouldn't be a good idea?

bigflow/deploy.py
```diff
@@ -35,18 +38,76 @@ def tag_image(image_id: str, repository: str, tag: str) -> str:
...
-        docker_repository: str,
-        auth_method: AuthorizationType = AuthorizationType.LOCAL_ACCOUNT,
-        vault_endpoint: T.Optional[str] = None,
-        vault_secret: T.Optional[str] = None,
+    image_tar_path: str,
+    docker_repository: str,
+    auth_method: AuthorizationType = AuthorizationType.LOCAL_ACCOUNT,
+    vault_endpoint: T.Optional[str] = None,
+    vault_secret: T.Optional[str] = None,
 ) -> str:
```
iwko, line 46: this function returns nothing

bigflow/build/operate.py
```diff
@@ -200,9 +276,15 @@ def build_project(
     project_spec: BigflowProjectSpec,
     start_time: str,
     workflow_id: typing.Optional[str] = None,
+    export_image_tar: bool | None = None,
```
TSienki, line 279: General comment - there is inconsistency in optional type hints. In some places the old form (Optional[T]) is used, in some the new (Optional | T). Look at the previous line. 
Shouldn't we choose one option?

TSienki: Especially in the context of one function.

bigflow/deploy.py
```diff
@@ -35,18 +38,76 @@ def tag_image(image_id: str, repository: str, tag: str) -> str:
...
+    if image_tar_path.endswith(".toml"):
```
iwko, line 47: If i understand correctly sometimes it is in fact path to tar_image but in other cases it is path not to image but to project_spec?

bigflow/deploy.py
```diff
@@ -35,18 +38,76 @@ def tag_image(image_id: str, repository: str, tag: str) -> str:
...
+    )
+
+
+def deploy_docker_image_from_local_repo(
+    image_info_path: str,
+    docker_repository: str,
+    auth_method: AuthorizationType = AuthorizationType.LOCAL_ACCOUNT,
+    vault_endpoint: str | None = None,
+    vault_secret: str | None = None,
+):
```
iwko, line 66: no return type hint

bigflow/cli.py
```diff
@@ -509,37 +543,66 @@ def _cli_deploy_dags(args):
...
-    files = glob1(".image", "*-*.tar")
-    if files:
-        return os.path.join(".image", files[0])
+
+    logger.debug("Scan folder .image")
+    if not os.path.isdir(".image"):
+        raise ValueError("Directory .image does not exist")
+
+    for f in os.listdir(".image"):
+        logger.debug("Foud file %s", f)
```
artnowo-alle, line 569: Found

bigflow/deploy.py
```diff
@@ -35,18 +38,76 @@ def tag_image(image_id: str, repository: str, tag: str) -> str:
...
+    deploy_fn(
+        image_tar_path,
+        docker_repository,
+        auth_method,
+        vault_endpoint,
+        vault_secret,
+    )
+
+
+def deploy_docker_image_from_local_repo(
```
iwko, line 60: this method is very similar to `_deploy_image_loaded_to_local_registry` can not we scale them?

bigflow/build/operate.py
```diff
@@ -72,14 +79,57 @@ def _export_docker_image_to_file(tag: str, target_dir: Path, version: str):
...
+        logger.debug("Authenticate to docker registry")
+        bigflow.deploy.authenticate_to_registry(
+            auth_method=cache_params.auth_method or bigflow.deploy.AuthorizationType.LOCAL_ACCOUNT,
+            vault_endpoint=cache_params.vault_endpoint,
+            vault_secret=cache_params.vault_secret,
+        )
+
+        image = cache_params.cache_from_image or ""
+        if cache_params.cache_from_version:
+            image += f",{project_spec.docker_repository}:{cache_params.cache_from_version}"
```
iwko, line 102: Can we move those two lines to some method with a significant name?

bigflow/cli.py
```diff
@@ -318,15 +320,41 @@ def _add_build_dags_parser_arguments(parser):
...
+    )
+    parser.set_defaults(export_image_tar=None)
+
+    parser.add_argument(
+        '--cache-from-image', dest='cache_from_image',
+        help="Docker images to consider as cache sources",
+    )
+    parser.add_argument(
+        '--cache-from-version', dest='cache_from_version',
+        help="Use previous version of the project as cache source",
```
iwko, line 340: If I understand correctly it doesn't have to be the previous version? It can be even a very old version from two years?

bigflow/cli.py
```diff
@@ -318,15 +320,41 @@ def _add_build_dags_parser_arguments(parser):
...
+    )
+    parser.add_argument(
+        '--no-export-image-tar', dest='export_image_tar', action='store_false',
+        help="Don't epxort built docker image into .tar file (keep image in local docker registry)",
+    )
+    parser.set_defaults(export_image_tar=None)
+
+    parser.add_argument(
+        '--cache-from-image', dest='cache_from_image',
+        help="Docker images to consider as cache sources",
```
iwko, line 336: Docker images or one single image?

bigflow/cli.py
anjensan: For now yest. It contains '-c'/'--config' option, which is used to resolve 'vault' token/key from 'deployment_config.py'

bigflow/cli.py
anjensan: Now missing value 'vault_secret' is ignored (honestly not sure why)
https://github.com/allegro/bigflow/blob/f2d9008b1e1eb653e7e8035a0a0b0cc623be0c9f/bigflow/cli.py#L498
this PR just doesn't change that

bigflow/build/spec.py
anjensan: No. Current behaviour is to export tar image, it should remain as default (at lest for the next minor version).

bigflow/deploy.py
anjensan: No, it might be a path to 'image**.toml' file, which is a placeholder for an information about build.
This mostly exists to mimic old behaviour of bigflow "as much as possible".
For example now it is possible to build image, modify some local files (which result in changed version) and then deploy that image with "previous version". That is because now version is parsed from a file name, not from current project version. That change tries to keep existing behaviour.

bigflow/cli.py
anjensan: Yes, It may be any version.

bigflow/build/operate.py
```diff
@@ -72,14 +79,57 @@ def _export_docker_image_to_file(tag: str, target_dir: Path, version: str):
...
+    cache_from_image: str | None
 
 
 def build_image(
     project_spec: BigflowProjectSpec,
+    export_image_tar: bool | None = None,
+    cache_params: BuildImageCacheParams | None = None,
 ):
+    if export_image_tar is None:
+        export_image_tar = project_spec.export_image_tar
```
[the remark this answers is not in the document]
anjensan: The idea is that we can disable exporting to tar on all CIs (via our custom github action) without requiring to change any existing build configs pipelines (so users still have a choice to use or not to use tar files locally).

bigflow/cli.py
anjensan: done

bigflow/cli.py
anjensan: fixed

bigflow/build/operate.py
anjensan: Sure, why not. Refactored.

bigflow/deploy.py
anjensan: fixed

bigflow/deploy.py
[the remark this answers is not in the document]
anjensan: They are not intended to be used outside of 'deploy.py'.
if you are hinting at '_prefix' - I've added it.

bigflow/deploy.py
anjensan: fixed

bigflow/deploy.py
anjensan: Refactored.

bigflow/build/operate.py
anjensan: IMO it's not very beneficial to move those 4 lines into a separate function (adds 4 more lines + extra view jump for a reader). But I'm not against this change.

Moved this into a new function.

bigflow/build/operate.py
```diff
@@ -72,14 +79,57 @@ def _export_docker_image_to_file(tag: str, target_dir: Path, version: str):
...
+            image += f",{project_spec.docker_repository}:{cache_params.cache_from_version}"
+            image = image.lstrip(",")
+
+        logger.debug("Add --cache-from=%s to `docker build`", image)
+        cmd.extend(["--cache-from", image])
+
+        logger.debug("Enable buildkit inline cache")
+
+    # noop when building backend is not a buildkit
+    cmd.extend(["--build-arg", "BUILDKIT_INLINE_CACHE=1"])
```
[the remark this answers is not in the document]
anjensan: Sure, logging should be aligned. Fixed.

bigflow/cli.py
anjensan: fixed

bigflow/build/operate.py
```diff
@@ -72,14 +79,64 @@ def _export_docker_image_to_file(tag: str, target_dir: Path, version: str):
...
+) -> str:
+    image = cache_params.cache_from_image or ""
+    if cache_params.cache_from_version:
+        image += f",{project_spec.docker_repository}:{cache_params.cache_from_version}"
+        image = image.lstrip(",")
+    return image
+
+
+@dataclass()
+class BuildImageCacheParams:
```
beczkowb, line 124: It would be nice to have static factories because if you don't know bigflow, you need to know that only specific combinations of these params make sense.

For example:

```
def cache_local(version_or_image: str) -> BuildImageCacheParams: ...
def cache_google_container_registry(version_or_image: str, vault_endpoint: str, vault_secret: str) -> BuildImageCacheParams: ...
```

bigflow/cli.py
iwko: 👍

bigflow/cli.py
iwko: 👍

bigflow/deploy.py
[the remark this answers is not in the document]
iwko: > if you are hinting at '_prefix' - I've added it.

exactly what I meant :D

bigflow/build/operate.py
iwko: 👍

bigflow/cli.py
```diff
@@ -318,15 +320,41 @@ def _add_build_dags_parser_arguments(parser):
...
+def _add_build_image_parser_arguments(parser: argparse.ArgumentParser):
+    parser.add_argument(
+        '--export-image-tar', dest='export_image_tar', action='store_true',
+        help="Export built docker image into .tar file",
+    )
+    parser.add_argument(
+        '--no-export-image-tar', dest='export_image_tar', action='store_false',
+        help="Don't epxort built docker image into .tar file (keep image in local docker registry)",
+    )
+    parser.set_defaults(export_image_tar=None)
```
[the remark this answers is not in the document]
artnowo-alle: The linked comment makes sense, however, I still have a concern with two contradictory switches. Add a note that they both overwrite what's in project spec and note what happens if none/both are used.

artnowo-alle:
There is an argument in deploy image command:
```
def _add_deploy_image_parser_arguments(parser):
    parser.add_argument('-i', '--image-tar-path',
                        type=str,
                        help='Path to a Docker image file. The file name must contain version number with the following naming schema: image-{version}.tar')
```
Shouldn't there be a no-tar equivalent?

bigflow/build/operate.py
anjensan: Sure, it's good to have consistent type-hints in a single declaration, I've just missed that (fixed).

In a wider sense - previously we used only "proper" annotations for "python 3.7" (`Option`, `Union` etc).
Now a "new-style" annotation from newer version of python are used, but old code remains the same, so we have some "mix". IMO there is no need to change hints immediately, but for new code I prefer "new style" where it is possible.

Also note then new-style `X | None`, `list[X]` are working only because of `from __future__ import annotations` declaration, which instructs python to keep annotations as a strings. So it is possible to write any "not working" expression if it may be parsed into correct python AST. However tools (at least mypy & pyright/pylance) know new ways of declaring type-hints and use them correctly.

bigflow/build/operate.py
anjensan: IMO such factories doesn't make big sense because they are not improving our CLI.
Adding them just add some `if .. else` branching into `cli.py` (the only place where `BuildImageCacheParams` is constructed).

Also note that now all options may be passed to `bigflow build` and unused are just ignored (like `--vault-endpoint` may be always passed even when auth type is "LOCAL").

anjensan:
> There is an argument in deploy image command:
> 
> ```
> def _add_deploy_image_parser_arguments(parser):
>     parser.add_argument('-i', '--image-tar-path',
>                         type=str,
>                         help='Path to a Docker image file. The file name must contain version number with the following naming schema: image-{version}.tar')
> ```
> 
> Shouldn't there be a no-tar equivalent?

I don't think so. Now "no-tar-ref-file" might be passed as "--image-tar-path" argument, but this is more like a side-effect (or a hidden feature).

IMO explicit pointing to a tar is not quite useful, especially for a new "no tar mode".
So I'd like not add a new option unless someone will explicitly request it.

artnowo-alle:
> I don't think so. Now "no-tar-ref-file" might be passed as "--image-tar-path" argument, but this is more like a side-effect (or a hidden feature).
> 
> IMO explicit pointing to a tar is not quite useful, especially for a new "no tar mode". So I'd like not add a new option unless someone will explicitly request it.

 Ok, so if you don't use tars you just don't pass the `--image-tar-file` arg and everything works correctly. Am I right?

bigflow/cli.py
iwko: So maybe I Would change `help` section of this method to `"Use older version of the project as cache source",`. For me "previous version" suggests only one last version

iwko (dismissed):

anjensan:
> Ok, so if you don't use tars you just don't pass the --image-tar-file arg and everything works correctly. Am I right?

Yes. Use-case for this parameter - two-phase deploying. When on one machine we run 'bf build-image' and on the second 'bf deploy-image --image-tar-file'. This just is not applicable to new feature, as local docker registry is used.

anjensan:
Also added parameter '--git-commit' to command 'bigflow project-version'.
It allows to look at the project version for a specified commit, like:
```bash
bigflow project-version --git-commit HEAD~2
```

PR #317, opened 2022-03-28
FST-163 bigflow docs clarifications and fixes

artnowo-alle:


docs/project_structure_and_build.md
```diff
@@ -66,8 +66,10 @@ project_dir/
     pyproject.toml
 ```
 
-Let us start with the `project_package`. It's the Python package which contains the processing logic of your workflows.
-It also contains `Workflow` objects, which arranges parts of your processing logic into
+Let us start with the `project_package`. It's the Python package which contains the processing logic of your workflows. BigFlow looks up for project root package in the relative path equal to `project_setup.PROJECT_NAME` or `--project-package` option value (`run` command only). Within the root package, it automatically discovers packages excluding directories starting with `test` prefix. Within the root package BigFlow acquires workflows by importing all modules and looking up for `bigflow.Workflow` instances.
```
grzegorzwitkowski, line 69: > excluding directories starting with `test` prefix
isn't it about files with test prefix? not directories

artnowo-alle:
```
def discover_project_packages(project_dir: Path):
    ret = setuptools.find_packages(where=project_dir, exclude=["test.*", "test"])
    logger.info(
        "Automatically discovered %d packages: \n%s",
        len(ret),
        "\n".join(map(" - %s".__mod__, ret)),
    )
    return ret
```

and `setuptools.find()` docstring says:

```
'exclude' is a sequence of package names to exclude; '*' can be used as a wildcard in the names, such that 'foo.*' will exclude all subpackages of 'foo' (but not 'foo' itself)
```

so...I suppose both of us are right. Good catch, will change it.

PR #327, opened 2022-10-28
* removing deprecated code and dependencies

beczkowb:
* removing deprecated dataproc and log modules with their dependencies
* adding pip and pip-tools constraints in requirements.in
* turning requirements.txt into requirements.in and generating requirements.txt using pip-compile
* changing docker image chaching implementation - using BUILDKIT_INLINE_CACHE=1 only if cache properties are set
* setting grpcio-status as <=1.48.2 to avoid problems with pip-compile on protobuf package
* always installing typing-extensions>=3.7 to avoid clashes
* changing test docker images from python:3.7 to python:3.7.15-slim to reduce image size


bigflow/__init__.py
```diff
@@ -32,13 +32,3 @@ class Workflow(bigflow.workflow.Workflow): ...
 
 @public(class_alias=True)
 class Config(bigflow.configuration.Config): ...
-
-# proactively try to initialize bigflow-specific logging
```
artnowo-alle, line 36: Why is this deprecated, did it actually work?

bigflow/cli.py
```diff
@@ -898,10 +884,6 @@ def cli(raw_args) -> None:
         _cli_project_version(parsed_args)
     elif operation == 'release':
         _cli_release(parsed_args)
-    elif operation == 'logs':
```
artnowo-alle, line 901: Ditto

bigflow/_version.py
```diff
@@ -1,2 +1 @@
-__version__ = '1.4.2'
-
+__version__ = '1.4.3dev1'
```
artnowo-alle, line 1: As we removed two features here (logs command and dataproc support), I think it should be 1.5.0. 

BTW Did we announce earlier that those features are deprecated and will be removed?

beczkowb: True, I will change it to 1.5.0dev1.

Yes, these features have been deprecated since Feb 2022 (although there was no slack or email communication). 

We know that inside the company there isn't a single person using Dataproc support.

There are some cases of "using" (people add the logging extension, but have no idea how it can be used) the log extras.

I opt for deleting the logging stuff anyway. Removing it from a bigflow project after bumping it to 1.5.0 comes down to removing a single property (log_config) from workflows and removing the log extras from bigflow in requirements.in.

Because the log extras is easily detectable through github search, I can just find these projects and create pull requests removing the log extras.

wdyt?

artnowo-alle: Sounds good, do it as you wrote.

beczkowb: ok

artnowo-alle (approved):

PR #332, opened 2022-11-15
changing the release workflow, so it tries to do a release if the ci …

beczkowb:
…workflow was successful and the version file was changed

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -21,8 +21,8 @@ jobs:
             # Diff HEAD with the previous commit
             $diff = git diff --name-only HEAD^ HEAD
 
-            # Check if a file under docs/ or with the .md extension has changed (added, modified, deleted)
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -5,15 +5,32 @@ on:
...
+        - uses: actions/checkout@v2
+          with:
+            fetch-depth: 2
+        - shell: pwsh
+          id: check_file_changed
+          run: |
+            # Diff HEAD with the previous commit
+            $diff = git diff --name-only HEAD^ HEAD
+
+            # Check if a file under docs/ or with the .md extension has changed (added, modified, deleted)
```
artnowo-alle, line 24: The comment does not reflect the code

artnowo-alle (changes requested):

.github/workflows/release.yml
beczkowb: done

artnowo-alle (approved):

PR #333, opened 2022-11-15
release plan fix

beczkowb:


.github/workflows/release.yml
```diff
@@ -11,41 +11,43 @@ jobs:
...
-
-            # Set the output named "version_changed"
-            Write-Host "::set-output name=version_changed::$HasDiff"
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 2
+      - name: Version file check
+        shell: pwsh
```
grzegorzwitkowski, line 20: why powershell?

.github/workflows/release.yml
```diff
@@ -11,41 +11,43 @@ jobs:
...
+        id: check_file_changed
+        run: |
+          # Diff HEAD with the previous commit
+          $diff = git diff --name-only HEAD^ HEAD
```
grzegorzwitkowski, line 24: strange variable name `$diff` with `$` at the beginning. Why not just `diff`?

grzegorzwitkowski: OK, it's the syntax for variables in PowerShell

grzegorzwitkowski (approved):

PR #335, opened 2022-11-15
1.5.2.dev2 bumping jinja

beczkowb:


artnowo-alle (approved):

requirements.txt
```diff
@@ -172,7 +172,7 @@ iniconfig==1.1.1
     # via pytest
 jaraco-classes==3.2.3
     # via keyring
-jinja2==2.11.3
+jinja2==3.1.2
```
grzegorzwitkowski, line 175: here the version of jinja2 is strict, but in base.txt is given with a range

grzegorzwitkowski (approved):

PR #341, opened 2022-12-06
Removing dataflow pins which are deprecated

thegunner157:


beczkowb:
Also, I suggest bumping the bigflow._version in this PR (dev release)

PR #357, opened 2023-01-30
Enabled TLS certificate verification in get_vault_token()

artnowo-alle:


bigflow/cli.py
```diff
@@ -514,7 +530,7 @@ def _resolve_vault_endpoint(args):
 def _resolve_property(args, property_name, ignore_value_error=False):
     try:
         cli_atr = getattr(args, property_name)
-        if cli_atr:
+        if cli_atr or cli_atr is False:
```
iwko, line 533: Is else part now even reachable?

artnowo-alle: Yup, when you use `--vault-endpoint-verify false`

agnieszkarybak: `getattr` can return `None`, and then it reaches the else part

bigflow/_version.py
```diff
@@ -1 +1 @@
-__version__ = '1.5.4'
+__version__ = '1.6.0.dev1'
```
agnieszkarybak, line 1: should be changed to '1.6.0'?

agnieszkarybak (commented):
lgtm, but I'm a coauthor ;)

bigflow/_version.py
artnowo-alle: Will do on a release branch

agnieszkarybak (approved):

PR #360, opened 2023-02-23
Release branch 1.6.0

artnowo-alle:


bigflow/_version.py
```diff
@@ -1,2 +1 @@
-__version__ = '1.5.5dev2'
-
+__version__ = '1.7.0.dev1'
```
grzegorzwitkowski, line 1: hmm, doesn't match with the PR title `Release branch 1.6.0`

artnowo-alle: Yeah, but 1.6.0 is already released (I just tagged commit [e6b2c32](https://github.com/allegro/bigflow/pull/360/commits/e6b2c324298a5e0188a1fd1d335a2cecb3de170f)). If I would leave 1.6.0 here it'd try to release it again. What I want to get is to release 1.7.0.dev1 once it's merged to master.

.github/workflows/release.yml
```diff
@@ -5,16 +5,23 @@ on:
     branches: [ master ]
     types:
       - completed
+  workflow_dispatch:
+    inputs:
+      ref:
```
agnieszkarybak, line 10: Do we need this input? Do we need to run the workflow from one branch's HEAD but release the version from another commit?

artnowo-alle:
Actual change released in 1.6.0 is in PR #357

.github/workflows/release.yml
artnowo-alle: I think we might want to do this. I'm fine with dropping it now and adding later if needed, though

grzegorzwitkowski (approved):

agnieszkarybak (approved):

.github/workflows/release.yml
artnowo-alle: Agreed with Agnieszka to remove it. One can install a requirement from a specific commit, branch, or tag instead.

PR #366, opened 2023-04-26
Parse only known args

agnieszkarybak:


bigflow/cli.py
```diff
@@ -268,7 +268,7 @@ def _parse_args(project_name: Optional[str], args) -> Namespace:
 
     _create_codegen_parser(subparsers)
 
-    return parser.parse_args(args)
+    return parser.parse_known_args(args)[0]
```
grzegorzwitkowski, line 271: WDYT about making it more obvious that the goal here is to return only bigflow args without extra args?
```suggestion
    bigflow_only_args = parser.parse_known_args(args)[0]
    return bigflow_only_args
```

TSienki: Can we test it somehow?

agnieszkarybak: I'll add some tests once we decide we're good with the change at all :)

.github/workflows/ci.yml
```diff
@@ -35,7 +35,7 @@ jobs:
         curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
         sudo add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/ubuntu bionic test"
         sudo apt update
-        sudo apt install docker-ce
+        sudo apt install docker-ce containerd.io
```
artnowo-alle, line 38: What is it needed for?

test/cli/test_cli.py
```diff
@@ -727,6 +727,27 @@ def test_should_call_cli_build_commands_with_NOW_and_ALL(
...
+    @mock.patch('bigflow.build.operate.build_project')
+    @mock.patch('bigflow.build.spec.read_project_spec')
+    @freezegun.freeze_time("2001-02-03 15:00:00")
+    def test_should_call_cli_build_with_additional_params(
+        self,
+        read_project_spec: mock.Mock,
+        build_project_mock: mock.Mock,
+    ):
+        # when
+        cli(['build','--workflow', 'ALL', '--start-time', 'NOW', '--additional-param', 'param'])
```
grzegorzwitkowski, line 739: Both test look almost the same, despite different arguments being used. WDYT about using `@pytest.mark.parametrize`?

agnieszkarybak: I actually did that accidentally, look now

grzegorzwitkowski (approved):

PR #371, opened 2023-05-24
Add docs regarding retrieving logs from Log Explorer

thegunner157:


agnieszkarybak (commented):
I would add the following information:
- how to retrieve worker logs when airflow dag fails (with example filters for k8s workers)
- how to retrieve logs for a failed beam job (with examples of filters)

docs/monitoring.md
```diff
@@ -96,6 +96,54 @@ workflow = bigflow.Workflow(
...
+
+![Dataflow filter](./images/Beam_select_logs.png)
+
+Then we can filter out proper information similarly like in Log Explorer. We can filter out severity of messages and type
+whether we want job logs or workers logs.
+
+![Dataflow logs details](./images/Beam_filter_logs.png)
+
+However, very often worker logs doesn't show all the details. Very useful is opening them via Log Explorer and then delete 
+filter there and add only job id. After that we can find all possible information which we want.
```
agnieszkarybak, line 143: Very useful is opening them via Log Explorer and then delete filter there and add only job id -> Very useful is opening them via Log Explorer and then deleting all the filters and adding only the job id

agnieszkarybak (approved):

PR #372, opened 2023-06-12
FST-921 impersonated creds function

alicja-nowak-allegro:


artnowo-alle:
Please fix PR title so that it matches ticket ID format

bigflow/bigquery/dataset_configuration.py
```diff
@@ -145,26 +121,25 @@ def resolve_tables_labels(self, env: str = None) -> Dict[str, Dict[str, str]]:
...
-    #
-    #     creds, pid = google.auth.default()
-    #     credentials = google.auth.impersonated_credentials.Credentials(
-    #         source_credentials=creds,
-    #         target_principal=service_account,
-    #         target_scopes=target_scopes,
-    #     )
-    #     return credentials
+    @staticmethod
+    def _credentials_for_impersonate_service_account(
```
artnowo-alle, line 132: Didn't we want to make it public so that anyone can use it?

bigflow/bigquery/dataset_configuration.py
```diff
@@ -67,38 +66,16 @@ def add_configuration(self,
...
-                    ).__dict__
-                    )
-
-        # if passed_credentials is None:
-        #     creds = self.credentials
-        # else:
-        #     creds = passed_credentials
+
+        resolved_impersonate_service_account = self.resolve_impersonate_service_account(env)
+        resolved_credentials = self._credentials_for_impersonate_service_account(resolved_impersonate_service_account)
```
artnowo-alle, line 78: I don't really get those two lines. Could you please elaborate on what does the first resolution do?

artnowo-alle (approved):

bigflow/bigquery/dataset_configuration.py
alicja-nowak-allegro: I wanted to pass in DatasetConfig constructor impersonate_service_account so it would be the same as in config responsible for Dataflow workflow. InteractiveDatasetManager requires credentials as google.auth.credentials.Credentials so in 77 I resolve impersonate_service_account from config for given env (same as every other field in DatasetConfig's constructor) and in 78 I create credentials (type google.auth.credentials.Credentials) out of SA for specific environment.

PR #374, opened 2023-06-26
2023-06 bump dependencies

artnowo-alle:


requirements.txt
```diff
@@ -1,60 +1,60 @@
 #
-# This file is autogenerated by pip-compile with python 3.8
-# To update, run:
+# This file is autogenerated by pip-compile with Python 3.7
```
agnieszkarybak, line 2: It would be best if you used Python 3.8 as that's the currently supported version

artnowo-alle: Done

agnieszkarybak (approved):

PR #381, opened 2024-04-10
BQ write and query jobs labeling

artnowo-alle:


bigflow/bigquery/dataset_configuration.py
```diff
@@ -16,14 +16,16 @@ def __init__(self,
                  is_master: bool = True,
                  is_default: bool = True,
                  tables_labels: Dict[str, Dict[str, str]] = None,
-                 dataset_labels: Dict[str, str] = None):
+                 dataset_labels: Dict[str, str] = None,
+                 job_labels: Dict[str, str] = None):
```
grzegorzwitkowski, line 20: Why not an empty list as a default param value?

bigflow/bigquery/dataset_configuration.py
```diff
@@ -36,7 +38,8 @@ def add_configuration(self,
                           properties: dict = None,
                           is_default: bool = False,
                           tables_labels: Dict[str, Dict[str, str]] = None,
-                          dataset_labels: Dict[str, str] = None):
+                          dataset_labels: Dict[str, str] = None,
+                          job_labels: Dict[str, str] = None):
```
grzegorzwitkowski, line 42: Wouldn't it be better to use empty dicts as defaults?

bigflow/bigquery/dataset_manager.py
```diff
@@ -285,6 +287,9 @@ def write(self, table_id: str, sql: str, mode: str) -> 'google.cloud.bigquery.ta
         job_config.destination = table_id
         job_config.write_disposition = mode
 
+        if self.job_labels:
+            job_config.labels = self.job_labels
```
grzegorzwitkowski, line 291: You could skip such if statements in many places if job_labels would be initialized with empty dict, not None

grzegorzwitkowski:
Probably the change lacks documentation

grzegorzwitkowski (approved):

